PR #311, opened 2019-12-16
Chant Audio player

jacobsanz97:
A “Play Audio” button with an ID is added to the bottom of the "chant-record.template.html" template.
The Backbone Marionette View "ChantRecordView.js" is modified to trigger an event when this button is pressed, and calls the mainPlay function (see below).

A number of functions are then added to "ChantRecordView.js":

mainPlay - The function interacts with the DOM to extract volpiano-syllable objects from the currently expanded panel. It then calls the parse_volpiano function on an array of these objects, and passes them to the volpiano2midi player.

get_vowel - The function loops through the input text, until it find a vowel character. Upon finding a vowel (the first vowel in the text), the program returns an integer associated with the program change which needs to occur to produce this vowel. For example: a corresponds to 0, e to 1, and u to 4. If no vowels are found, it returns 0.

parse_volpiano - The array of volpiano-syllable objects fetched from the DOM is iterated over. Each element’s textContent is retrieved, as is split into the volpiano note and the syllable using the “--” token. The separated parts of the string are made into an array (as a 2-tuple), and this tuple is pushed to a final array of 2-tuples.

volpiano2midi - A dictionary is constructed, mapping Volpiano notes to numerical MIDI pitches. For example, the character ‘a’ maps to a pitch of 45.  Then, MIDI.js is setup using the generated SoundFont, and the volume of the first channel is set. A loop proceeds to iterate through the inputted volpiano-syllable array, and for each cell, obtains the pitches and vowel of the syllable, and performs a program change to load the necessary vowel SoundFont. A second loop iterates over each note in the cell, converting them to lowercase, and checking if they are contained in the dictionary. If the note is contained in the dictionary, it is played for the duration of time indicated by the input. If it is not contained in the dictionary, it is simply skipped.

CustomSplit - Home-brewed version of JavaScript's str.split function. This version can handle null/undefined inputs, so is more robust.

strLength - As above, a home-brewed method to replace the string.length property, which can’t handle null inputs.

dynamicallyLoadScript - Creates a script DOM node, and sets its src to the input URL, and adds it to the end of the head section of the page. Used to include the MIDI.js player files.


fujinaga:
There should be a "stop" function.
The "Play Audio" button should be deactivated until the chant finishes or it is stopped.
The "Play Audio" button should be absent when there's no Volpiano.

napulen:
The code of this PR has been merged already, so I am writing these as separate issues.

napulen:
Moved to #312

fujinaga:
Thank you.

Issue #396, opened 2022-05-09
vagrant up gets stuck on Windows environments


napulen:
When running the 
```
vagrant up
```

routine on a freshly cloned Windows repo, the script gets stuck at the following line:

```
default: Processing triggers for man-db (2.8.3-2ubuntu0.1) ...
```

This seems to happen in, both, my Windows 10 environment and @eliwaksbaum 's

PR #397, opened 2022-05-09
fix hanging on man-db tasks when building on windows

eliwaksbaum:
fix for #396, remove man-db package in provision

### DIFF
--- a/etc/provision/setup-privileged.sh
+++ b/etc/provision/setup-privileged.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+#fix hanging for windows environments
+sudo apt-get remove -y --purge man-db
+
 echo "==== Enable python dev, venv, and pip ===="
 apt-get update
 apt-get install -y python3-dev python3-venv python3-pip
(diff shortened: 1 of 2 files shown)


eliwaksbaum:
small regex fix stops static pages from crashing on windows

napulen (changes requested):
The fix for the provisioning looks great!

Regarding the url regex, I am concerned that leaving an open-ended string (no `$` check) would potentially open this code to security vulnerabilities (people adding extra characters at the url and passing the validation). I understand that this is a problem related to the Windows CRLF syntax (`\n\r`)?

How about we do an optional return carriage symbol before the end of line `$`? That'd probably work for *nix/Windows while not allowing arbitrary urls.

napulen:
I think something like this would work? (Did not test)

```
r"^#([A-Za-z0-9 _-]+)\r?$
```

Can you try moving in this direction for this fix, @eliwaksbaum?

napulen (approved):
Nice!

Issue #399, opened 2022-05-11
Add an "Edit" button in Folio Mapping interface


napulen:
Once a folio mapping has been submitted, users should be able to edit it.

dchiller:
Update: draft pull request #401

dchiller:
@napulen Should we close this issue now that we've merged?

napulen:
Yes!

PR #401, opened 2022-05-13
Add edit folio mapping functionality

dchiller:
On admin/map_folios menu, provide a complete list of manuscripts with folio mapping status and option to either map folios or edit map folios.

Updates view logic for edit folio mapping: folio mapping is populated with the last saved folio mapping, rather than default.


napulen:
Fixes #399.

napulen (approved):

napulen:
The PR is ready to move away from `draft` status. Once that's done, merging it into the main codebase.

PR #403, opened 2022-05-16
Updating documentation to current tech stack.

### LINKED ISSUE #400: Update documentation to account for gunicorn serving Cantus Ultimus (rather than `runserver`)
The documentation mentions that the django webapp is served via `python runserver`.

In the past, that used to be correct. However, in recent versions of the vagrant provisioning we have replaced that with `nginx` + `gunicorn` + `systemd` config files.

Also, the app is being served on port `8001`, instead of `8000`.

napulen:
Fixes #400

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,9 +37,32 @@ $ vagrant plugin install vagrant-openstack-provider
 
 # Set up the VM (this will take a while)
 $ vagrant up
+```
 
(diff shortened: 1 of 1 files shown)


napulen:
Using [mermaid ](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/) to provide some diagrams.

dchiller (approved):
Looks good to me!

Issue #398, opened 2022-05-09
Adding new django tests to run after a Cantus Ultimus instance first launches


napulen:
We can use the django tutorial for automated tests as a reference: https://docs.djangoproject.com/en/3.1/intro/tutorial05/

Things that would make for good tests:

- Make sure that making an http request to all of the static pages (e.g., `home`, `about`, `team`) returns some "proper" HTML content. (Left to decide what "proper HTML content" means in this case)
- Make sure that importing the chants of a well-known manuscript from Cantus DB results in the expected number of chants

eliwaksbaum:
@napulen is there a problem with the existing tests? I've been working on trying to make those run

napulen:
@eliwaksbaum I forgot about them! 

Some of them are obsolete. The ones under `mei_conversion` and `test_gamera_xml2mei` should be ignored/removed at this point. I don't think we'll reuse that MEI workflow anymore.

The one under `management` is relevant but outdated.

The `models` tests seem good, worth adapting. Same for `views`. I'd say, in order of priority:

1. `models`
2. `views`
3. `management`

napulen:
Thanks for noticing this!

eliwaksbaum:
```
System check identified 1 issue (0 silenced).
.F.....Deleting old concordances data...
Successfully imported 12 concordances into database.
Waiting for Solr to finish...
Done.
.....................Not Found: /chant/2f63f986449349769d7a313e0fc6edb3/
......Not Found: /concordance/2f63f986449349769d7a313e0fc6edb3/
...Not Found: /chant/2f63f986449349769d7a313e0fc6edb3/
..Not Found: /folio-set/manuscript/3/66666666666/
....Not Found: /manuscript/2f63f986449349769d7a313e0fc6edb3/
.Not Found: /manuscript/2/
...
======================================================================
FAIL: test_expand_genre (cantusdata.test.helpers.test_expandr.ExpandrFunctionsTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/code/public/cantusdata/test/helpers/test_expandr.py", line 44, in test_expand_genre
    self.assertEqual(expandr.expand_genre("Pr"), "Prefatio")
AssertionError: 'Prefatio (when written out and notated)' != 'Prefatio'
- Prefatio (when written out and notated)
+ Prefatio

----------------------------------------------------------------------
Ran 47 tests in 54.166s
```
[...]

PR #406, opened 2022-05-19
Unit tests

eliwaksbaum:
allow the existing unit tests to be run on the current vagrant build, see #398


public/cantusdata/test/management/commands/test_import_all_data.py
```diff
@@ -9,7 +9,8 @@ def setUp(self):
         pass
 
     def test_import_all_data(self):
-        call_command('import_data', 'test/test_chants.csv', '1', all=True)
+        call_command('import_data', 'concordances')
+        call_command('import_data', 'manuscripts')
```
napulen, line 13: The original code seems to be importing chants, whereas the new one `concordances` and `manuscripts`. Is that the expected behaviour? That is, will the unit tests pass/fail accordingly?

public/cantusdata/helpers/expandr.py
```diff
@@ -86,7 +86,9 @@ class PositionExpander(object):
     position_data_base = None
 
     def __init__(self):
-        self.csv_file = csv.DictReader(open("data_dumps/position_names.csv"))
+        cantusdata_path = os.path.split(os.path.dirname(__file__))[0]
+        public_path = os.path.split(cantusdata_path)[0]
+        self.csv_file = csv.DictReader(open(os.path.join(public_path, "data_dumps", "position_names.csv")))
```
napulen, line 91: If I understand this correctly (i.e., getting the absolute path of django project), this is available through the `settings` file of the project. You could import that path in the `settings.py` of the django project and avoid a redundant call to `os.path.dirname(__file__)`. 

Please check the `settings.py` and let us know if that's an appropriate import for this information.

napulen (commented):
See comments.

public/cantusdata/test/management/commands/test_import_all_data.py
eliwaksbaum: The other tests are not using the data that is imported here. The point of this specific test is to test the ability to import data and as soon as it's done it deletes everything. I changed it from chants to concordances and manuscripts as you can't import chants independently anymore, nor does it seem like you can provide a directory as an argument to import stuff from.

public/cantusdata/helpers/expandr.py
eliwaksbaum: yup. i'll make the fix

public/cantusdata/test/management/commands/test_import_all_data.py
napulen: Yes, you can't import chants independently anymore. If the purpose is importing data and deleting it right after, then this is perfect. Thanks!

napulen:
Looking good now! Thanks!

napulen (approved):

PR #408, opened 2022-05-20
library speed tests

### LINKED ISSUE #394: Measure the networking speed of source libraries

eliwaksbaum:
Fixes #394


napulen (changes requested):
All good! Just waiting for the new tests based on Andrew's feedback.

napulen:
For some reason, the Github automation doesn't automatically close an issue on merge unless you explicitly write a word like `Fixe[d|s] #0000`. Please always include such a verbal cue before the issue number.

eliwaksbaum:
I'm discovering all sorts of interesting things about requesting IIIF images and these tests are not up to snuff. I'll reopen once I figure out what exactly we want to do here. For example, one issue that comes up is the transformations not working since something in the uri isn't right, so we just get the same thumbnail over and over. Some test in here should be looking for that. There's more to knowing if a manifest will work with our system than whether we get a 200 response back

napulen:
This is starting to sound like we need a IIIF module in the codebase, for parsing, error-handling, testing, etc.

PR #420, opened 2022-05-26
iiif module

### LINKED ISSUE #394: Measure the networking speed of source libraries

### LINKED ISSUE #415: more folios than canvasses crashes map_folios
```
if mapped_folios == 0:
            for idx, folio in enumerate(folios):
                uris_objs[idx]["folio"] = folio
```

in implementing this default mapping, we assume there at least as many canvasses as folios, which is not always the case, resulting in index out of bounds errors

eliwaksbaum:
fixes #415
fixes #394


napulen:
Let's hold a bit on this PR until we get the new `main` thing sorted out.

It's likely that this PR will need to go to a different branch with an incompatible version history.

eliwaksbaum:
i can also wait for the unified main, pull that, and then apply these changes from scratch on top of that history

eliwaksbaum:
it'll just be copying in one folder and changing one line

PR #423, opened 2022-05-26
IIIF and testing

### LINKED ISSUE #394: Measure the networking speed of source libraries

### LINKED ISSUE #415: more folios than canvasses crashes map_folios

### LINKED ISSUE #417: Run unittests with github actions
The Cantus Ultimus repo needs some Continuous Integration.

For starters, what I think we need:
- On `push` and PRs to the `main` branch, we want to run a check for `black` formatting all across the python codebase
- On `push` and PRs to the `main` branch, we want it to run unit tests and pass

### LINKED ISSUE #424: 500 Error gives 404 Error
Nginx tries to do a redirect when there's a 500 error to a page that doesn't exist, giving a 404 error. Ideally, the app service on the production site would never cause a server error, but still seems non-ideal

### LINKED ISSUE #429: nginx image relies on app image
There's no official support for specifying what order images get built in when running docker compose (which kind of implies to me that images shouldn't rely on each other?), but right now the app image needs to be built before the nginx image. I see three solutions:
1. try something hacky
2. manually build the app image first
3. redo the nginx dockerfile to build the django app (maybe a minamilzed version), copy over the files, and discard the image, the same way we do the node frontend thing now

I think 3 is probably the way to go? I'll investigate next week

### LINKED ISSUE #433: Cannot scroll results in search pop-up
A possible issue with the search pop-up. You cannot scroll through search results in the search pop-up.
image
Occurs in both Chrome and Firefox.

eliwaksbaum:
fixes #415
fixes #394
fixes #417
fixes #429
fixes #433
fixes #424


eliwaksbaum:
that last one was an old commit that got lost somewhere. tests won't work without it because the request to solr was using an outdated path

napulen (commented):
Shouldn't this test case be under the `test`? 

```
folder app/public/cantusdata/iiif/test_iiif.py
```

eliwaksbaum:
We could. This is just an easy way to keep things separate for runtests.py. If this test is within the test folder, then to run all the internal tests you'd have to do runtests.py cantusdata.test.views, runtests.py cantusdata.test.models, etc. Right now, runtests.py cantusdata.test runs everything, and runtests.py cantusdata.iiif runs these manifest tests that rely on external libraries

napulen:
How about `cantusdata.test.core` and `cantusdata.test.extra`?

So something like this (or similar)...

```
cantusdata/test
├── __init__.py
├── core
│   ├── helpers
│   │   ├── __init__.py
│   │   ├── test_expandr.py
│   │   └── test_gamera_xml2mei
│   │       └── __init__.py
│   ├── management
│   │   ├── __init__.py
│   │   └── commands
│   │       ├── __init__.py
│   │       └── test_import_all_data.py
│   ├── mei_conversion
│   │   └── __init__.py
│   ├── models
│   │   ├── __init__.py
│   │   ├── test_chant.py
│   │   ├── test_concordance.py
│   │   ├── test_folio.py
│   │   ├── test_manuscript.py
│   │   └── test_plugin.py
│   └── views
│       ├── __init__.py
│       ├── test_browse.py
│       ├── test_chant.py
│       ├── test_chant_set.py
│       ├── test_concordance.py
│       ├── test_folio.py
│       ├── test_folio_set.py
│       ├── test_manuscript.py
│       └── test_search.py
├── extra
│   └── iiif
 |      └── ...
└── test_settings.py
```

PR #426, opened 2022-05-27
Modifies Solr synch signal to allow for concurrent db transactions

dchiller:
The prior instantiation of the signal would raise a set-related error
when multiple processes were making DB transations. Items would be added
to the SynchronizationSession objects _deletions and _additions attributes
during iteration (ie. when the execute method was called).

Adds an executing flag to the Solr synchronizer so that solr additions
and deletions will not be added to the set during iteration. Instead,
another SynchronizationSession will be created.

### DIFF
--- a/app/public/cantusdata/signals/solr_sync.py
+++ b/app/public/cantusdata/signals/solr_sync.py
@@ -48,8 +48,9 @@ class SolrSynchronizer(object):
     @contextlib.contextmanager
     def get_session(self):
         if self._session:
-            yield self._session
-            return
+            if not self._session._executing_flag:
+                yield self._session
+                return
 
         self._session = SynchronizationSession(self.solr_server_url)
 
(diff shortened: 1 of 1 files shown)


app/public/cantusdata/signals/solr_sync.py
```diff
@@ -103,7 +104,9 @@ def __init__(self, solr_server_url):
         self._deletions = set()
         self._additions = set()
 
+        self._executing_flag = False
     def execute(self):
+        self._executing_flag = True
```
dchiller, line 109: Make deep copies of _deletions and _additions so they are saved before the SynchronizationSession is overwritten in SolrSynchronizer.get_session().

app/public/cantusdata/signals/solr_sync.py
```diff
@@ -106,14 +106,16 @@ def __init__(self, solr_server_url):
 
         self._executing_flag = False
     def execute(self):
-        self._executing_flag = True
         conn = solr.SolrConnection(self.solr_server_url)
```
dchiller, line 109: Move this to below the executing_flag = True

dchiller:
@napulen Ok, I think this is good now. :)

PR #430, opened 2022-05-30
Configure database password through environment variables

### LINKED ISSUE #355: Protect database passwords through environment variables
Define environment variables for the database

```python
DATABASES = {
    "default": {
        "PASSWORD": "ThePassword",
     }
}
```

dchiller:
Postgres admin password is now created through environment variables
in the .env file. Deploying users can customize the file with a secure
password by locally changing this file.

Fixes #355


dchiller:
FYI: I only added information about this specific element to the README, thinking that there probably needs to be a more comprehensive edit of that doc for the whole dockerport.

.gitignore
```diff
@@ -1,4 +1,5 @@
 __init__.pyc
+.env
```
napulen, line 2: I think we do need `.env` on version control, which means, **not** in the `.gitignore`

docker-compose.yaml
```diff
@@ -14,9 +14,9 @@ services:
     volumes:
       - ./data/postgres:/var/lib/postgresql/data
     environment:
-      - POSTGRES_DB=cantus_db
-      - POSTGRES_USER=cantus_admin
-      - POSTGRES_PASSWORD=Pl4c3H0ld3r
+      - POSTGRES_DB=${POSTGRES_DB}
```
napulen, line 17: This only works iff there is an `.env` file, otherwise you set an empty password, right? Kind of important. Not sure if there's a better way to do this.

napulen (commented):
Willing to approve the method in `docker-compose.yaml`, pending we force/remind/encourage users to always verify they have a `.env` file. Or propose a more robust method.

docker-compose.yaml
dchiller: I did not test how this requirement makes itself known, but the docs say that it cannot be empty, and that this environment variable is required by the container. So, I think for deployment anywhere in Canada we should require that this be set and not left empty? Let me see if I can check how this is done -- whether it errors in an informative way, etc?

.gitignore
dchiller: Gotcha -- so was wondering the best way to do this if we want .env in version control but I don't want anyone to edit their own .env with the password and then push to the repo....I guess that's something thought that we would just expect to catch on  review?

docker-compose.yaml
dchiller: OK - so if you don't specify the password, the container build fails with:
image

I do agree that we want to force people to set a password and that we don't want to have a default password written in the codebase somewhere that could go to production without being modified or reviewed.

.gitignore
napulen: Aha, gotcha. Yes, we avoid that by blocking the `main` branch and requiring a review of every PR.

napulen: I don't think there is anything like a "read-only" file in version control, which would be the solution to this. Just making it more difficult for destructive commits to get in the codebase :)

dchiller: Perfect! Will make the change.

napulen (approved):

dchiller:
Fixes provided by #432

PR #436, opened 2022-06-01
Add healthcheck on postgres container to docker-compose.yaml

dchiller:
Changes to docker-compose.yaml mean that build of the app container
will wait until the postgres container has status "healthy". Postgres
container is determined healthy when pg_isready, the Postgres command
that tests whether server is accepting connections, returns a positive
result.


napulen (approved):
Looking good

PR #442, opened 2022-06-01
Remove unused commands

dchiller:
Removes the update_public_manuscripts.py command. 

Function allowed for the bulk setting of manuscripts to public via a csv upload. This should be completed in the future through the admin site as it currently exists or through an extension of the import_data functionality.


napulen:
This can stay open should we decide to clean some more files beyond `update_public_manuscripts`

napulen (approved):

PR #445, opened 2022-06-02
Upgrade to Django 4

dchiller:
- Upgrades to Django 4
- Upgrades to python10 in app container to support Django 4
- Runs `python manage.py migrate` in background during app build so site is available when existing data is indexed by solr.


dchiller:
TO DO:
- Some navigation around the site raises a "Session data corrupted" message in the app container -- no seeming loss of performance in the site, but investigating.

PR #449, opened 2022-06-02
Security warnings

### LINKED ISSUE #356: running check --deploy
Currently, this `python manage.py check --deploy` throws the following:

```python
System check identified some issues:

WARNINGS:
?: (security.W004) You have not set a value for the SECURE_HSTS_SECONDS setting. If your entire site is served only over SSL, you may want to consider setting a value and enabling HTTP Strict Transport Security. Be sure to read the documentation first; enabling HSTS carelessly can cause serious, irreversible problems.
?: (security.W008) Your SECURE_SSL_REDIRECT setting is not set to True. Unless your site should be available over both SSL and non-SSL connections, you may want to either set this setting True or configure a load balancer or reverse-proxy server to redirect all connections to HTTPS.
?: (security.W012) SESSION_COOKIE_SECURE is not set to True. Using a secure-only session cookie makes it more difficult for network traffic sniffers to hijack user sessions.
?: (security.W016) You have 'django.middleware.csrf.CsrfViewMiddleware' in your MIDDLEWARE, but you have not set CSRF_COOKIE_SECURE to True. Using a secure-only CSRF cookie makes it more difficult for network traffic sniffers to steal the CSRF token.
?: (security.W018) You should not have DEBUG set to True in deployment.
?: (security.W020) ALLOWED_HOSTS must not be empty in deployment.
cantusdata.Manuscript.plugins: (fields.W340) null has no effect on ManyToManyField.
```

All of these warning should pass (or we should be sure that they are not a problem) before deploying on production.

eliwaksbaum:
redo of #448

fixes #356

one warning remains, SECURE_SSL_REDIRECT is not set to true. "Unless your site should be available over both SSL and non-SSL connections, you may want to either set this setting True or configure a load balancer or reverse-proxy server to redirect all connections to HTTPS." But we do have a reverse-proxy that does just that, right? And setting this to true makes the site inaccessible. So I think we're ok.


napulen (approved):
Looks good

PR #451, opened 2022-06-02
Readme

### LINKED ISSUE #446: Update documentation for docker version
The current documentation accounts for a development/production workflow based on `Vagrant`.

It needs to be replaced to account for how things work now with `docker`.

eliwaksbaum:
fixes #446

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,30 +21,31 @@ git submodule update --init
 ```
 
 ## Launching the website
-We use Docker Compose to containerize each service and keep all our dependencies in order.
+We use [Docker Compose](https://docs.docker.com/compose/) to containerize each service and keep all our dependencies in order.
 
-### The .env file
-The build process relies on enviornment variables specified in the .env file in the root of the repository. Most importantly `POSTGRES_PASSWORD` must be set by uncommenting the appropriate line and adding a secure password.
+### The `.env` file
+The build process relies on environment variables specified in the `.env` file, which is located in the root of the repository. The most important variable here is `POSTGRES_PASSWORD`, which must be set by uncommenting it from the code, and setting it with a secure password.
 
-### Launch in development (Vagrant + Virtualbox)
+### Launch in development
 
-In the .env file, change `DEBUG=false` to `DEBUG=true`. This will turn on Django's debug mode, showing detailed traces when Django encounters an error, as well as turn off certain security settings that might stop you from accessing the site locally.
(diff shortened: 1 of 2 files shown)


napulen (approved):
This is great! Thank you @eliwaksbaum. I added some changes.

napulen:
I'll re-read before merging it.

dchiller:
What's the change to the nginx part of docker-compose.yaml?

dchiller (approved):
Love the readme!

napulen:
> Maybe one thing to add to the read-me is about that error where (if you are developing) you need to make sure the folders where the postgres db is persisted are deleted -- or else upon rebuild postgres will not initialize....

Yes, that's important

eliwaksbaum:
> Maybe one thing to add to the read-me is about that error where (if you are developing) you need to make sure the folders where the postgres db is persisted are deleted -- or else upon rebuild postgres will not initialize....

That happens when you change the password, right?

napulen:
> > Maybe one thing to add to the read-me is about that error where (if you are developing) you need to make sure the folders where the postgres db is persisted are deleted -- or else upon rebuild postgres will not initialize....
> 
> That happens when you change the password, right?

Yeah, I think the particular case I ran into was
- I forgot to set up the password on `.env` and launched the container
- Realized that I didn't set up a password, so I edited `.env`
- I re-built the containers and ran into this `postgres` issue

napulen:
I pushed a rough draft of the `postgres` issue

napulen:
I also re-read the complete `README` file and I am good to go.

Can you review/double-check/edit the `postgres` section? After that, I think we are good to go with this documentation 👍

eliwaksbaum:
Is removing the volume with the docker cli enough or do you have to actually delete the files on your machine?

napulen:
> Is removing the volume with the docker cli enough or do you have to actually delete the files on your machine?

I don't know for sure, but I can just add the `rm` instruction to the readme.

dchiller:
I have always just removed the files, independent of docker, but if removing the volume does the trick, looks great!

napulen:
The problem I had removing just the files is that I run into a weird `permission denied` issue, probably because something on the docker daemon is holding that folder.

napulen:
Ok, merging the current docs. Corrections can come in future PRs as usual.

PR #453, opened 2022-06-02
Upgrade to Django4

dchiller:


napulen:
@dchiller did this configuration result in a working set of containers?

dchiller:
Yes! Everything is working -- the only thing I encountered was the "Session data corrupted" error, but since Eli found that in the 3.2 version too, I think it is a separate issue.

napulen:
Awesome! So we can leave the `django4` branch lying there. Today, I noticed that the Alpha for `django==4.1.0` is out, so we could technically be early adopters of 4.1 and then we have full support until 4.2 comes out.

Not sure that it's worth it, because it is an "alpha" after all.

napulen:
So I'm still in favor of leaving this one here for keeping this conversation alive by the time `4.2` is out.

dchiller:
Closing....solved with #744

PR #457, opened 2022-06-02
Run initial migrate command in background.

dchiller:
Running the migrate command results in a complete reindexing of the
Solr backend. Running in background allows access to website while
indexing is completed.

### DIFF
--- a/app/django-config.sh
+++ b/app/django-config.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
 python manage.py makemigrations
-python manage.py migrate
+python manage.py migrate &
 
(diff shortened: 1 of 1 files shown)


napulen (commented):
The `solr` reindexing should only have the basic model templates, right? 

I don't see harm in this change, but I am curious on the time elapsed on this `makemigrations`. It shouldn't be much different from the one running on the background, is it?

dchiller:
You are correct, unless you are rebooting the docker container, but keeping the same database....in other words, what if we make a change to a view, but don't want to have to reimport everything or remap everything if the db is fine.

In that case, everything in the db is reindexed.

PR #459, opened 2022-06-03
Matching django versions in requirements

napulen:
Matching django versions in `app/public/requirements.txt` and `nginx/public/python/requirements.txt`


napulen:
Merging after either (or both) of your reviews (merge blocked otherwise).

Issue #150, opened 2015-03-16
504 Gateway Time Out on Dev Cantus


AFFogarty:
I'm getting Time Out whenever I try to search on dev-cantus.  This is not happening with production Cantus.

Example: http://dev-cantus.simssa.ca/search/?q=dom%20%20AND%20manuscript:%22CDN-Hsmu%20M2149.L4%22&sort=folio%20asc

AFFogarty:
Restarting Tomcat did not fix this.

AFFogarty:
Ok, this is definitely a Solr-related problem.  When I rebuild Solr for Dev-Cantus the issue is fixed for Dev-Cantus but the same issue suddenly appears on Cantus.  When I rebuild Solr for Cantus then the issue appears on Dev-Cantus.

ahankinson:
Did you change the data-dir?

ahwitz:
I don't know if this is related, but I was trying to look up something about the Salzinnes and realized that all the manuscript IDs are one greater in dev-cantus than in cantus.

ahankinson:
Has this been fixed? It should be the first priority to make sure this works correctly.

ahankinson:
Seems to be fixed now.

PR #463, opened 2022-06-03
Add intermediate page after import chant action

dchiller:
Allows for the bulk of the import chant command to complete
on a thread in the background called from the new intermediate
page view. Solves issue of import chant action timeout in
admin page.

Intermediate solve for #150


napulen:
This looks pretty cool! I may have a few questions, but I'll run it locally first.

PR #468, opened 2022-06-07
Fixes a django error when saving changes to manuscript objects

dchiller:
The manuscript model's save method was re-indexing chants in solr
even if chants had never been loaded. Although the error had no
impact on performance, the error is now avoided.


napulen (approved):
Nice!

Issue #478, opened 2022-06-07
Update IIIF manifests page 


dchiller:
Two spreadsheets with additional manifests:

- https://docs.google.com/spreadsheets/d/1bNaDkkweEpffpzSH2GkyIN5P_xWNaBqZ/edit#gid=1700575581
- https://drive.google.com/file/d/1owzNnXDHCkFT7FfgPkFCpeFqyU0ituTX/view?ts=60a27de2

napulen:
Yeah, we can get rid of it

dchiller:
Completed with #481

PR #481, opened 2022-06-07
import_data iiif

eliwaksbaum:
automatically sets the manifest_url field of the manuscript objects. with this command, do we still need a page to display the links a la #478?

before we merge, @napulen :
- What does `@transaction.atomic` do? should it tag the new function as well?
- Where should `manifests.csv` go?


eliwaksbaum:
would also require an update to the README i guess

napulen:
> would also require an update to the README i guess

yes

napulen:
> * Where should `manifests.csv` go?

In the past, everything has been under `data_dumps`. In lack of a better suggestion (because this is a really unique piece of information), I'd suggest `data_dumps`.

.env
```diff
@@ -7,4 +7,4 @@ PORT=8000
 #Postgres authentication variables
 POSTGRES_DB=cantus_db
 POSTGRES_USER=cantus_admin
-#POSTGRES_PASSWORD=**PROVIDE PASSWORD HERE**
+POSTGRES_PASSWORD=monkeybutt
```
dchiller, line 10: an accidental part of the commit?? 😜😜

app/public/cantusdata/management/commands/import_data.py
```diff
@@ -125,3 +130,16 @@ def import_chant_data(self, **options):
...
+
+    @transaction.atomic
+    def import_iiif_data(self):
+        with open(path.join(BASE_DIR, "data_dumps", "manifests.csv"), "r") as file:
+            csv = file.readlines()
+
+        for row in csv:
+            siglum_manifest = row.split(",")
+            qs = Manuscript.objects.filter(siglum=siglum_manifest[0].strip())
+            if len(qs) > 0:
```
dchiller, line 142: So this is catching the times that the query in line 141 doesn't yield anything (ie when `len(qs) = 0`)?

.env
napulen: I can edit this one out

eliwaksbaum: lol i knew i'd forget eventually. good catch

app/public/cantusdata/management/commands/import_data.py
eliwaksbaum: yeah exactly

napulen (changes requested):

napulen:
Cool!

PR #493, opened 2022-06-14
Adds celery asynchronous task queue

dchiller:


app/public/cantusdata/admin/admin.py
```diff
@@ -53,10 +54,11 @@ class ManuscriptAdmin(admin.ModelAdmin):
 
     @admin.action(description = "Imports the chants associated \
         with the selected manuscript(s)")
+
     def load_chants(self, request, queryset):
-        manuscript_ids = [str(manuscript.pk) for manuscript in queryset]
-        manuscript_ids_str = ','.join(manuscript_ids)
-        return HttpResponseRedirect(f'/admin/cantusdata/manuscript/load_chants/?manuscript_ids={manuscript_ids_str}')
+        manuscript_ids = [manuscript.pk for manuscript in queryset]
```
napulen, line 59: Missing `black` formatting?

rabbitmq/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM rabbitmq
```
napulen, line 1: Is this all this container needs?

napulen (changes requested):
In general, all files need consistent formatting. Fixable by running `black` on all.

napulen:
I haven't build the containers over this branch, got a bit swamped with thesis work. I'll try to make some time for this soon.

napulen:
But the changes make sense and look good :D

rabbitmq/Dockerfile
dchiller: Yes

dchiller: It uses some environment variables but those are in the env and docker-compose files

app/public/cantusdata/admin/admin.py
dchiller: I thought that github action that I wrote was supposed to take care of this but it looks like it didn't....I will check on that.

napulen: Oh, the gh-actions will `fail` the build if it's not formatted. 

An automated formatting would involve running `black` on gh-actions and then *committing* the changes. It can be done, but I don't think it is a good idea. Commits that modify the source code should always come from a human.

napulen: So far, what we have is a `black` "checker" :)

dchiller: makes sense. it passed that check though, so I still need to take a look since it should have failed.

napulen: can be a separate PR

napulen:
It seems we were missing formatting in a lot of places. Unfortunately, makes it harder to identify the changes related with celery, and our unit tests are failing, so there will be changes necessary on the celery portion. Sorry, this is on me for not pushing an issue for overall formatting before.

How about this:
- [ ] Revert last commit on this PR, go back to your proposed changes
- [ ] Make PR for fixing the gh-actions `--check`
- [ ] Make PR for overall formatting on existing codebase (no celery)
- [ ] Retake the formatting of this PR to get the formatting applied only to newer files

dchiller:
I think between this PR, #495, and #496

napulen:
Thanks for reviewing the formatting and getting the related PRs merged @dchiller

Back to the celery topic. It seems up to the latest commit, we are not passing the existing unit tests.

This is the error message:

image

Just based on the current committed files, we probably need an update in our gh-actions so that it injects the password for `RABBITMQ`, could this be the reason why it is failing the unit tests?

dchiller:
Yeah, I saw that...I am looking at it.

dchiller:
Test passed! @napulen 

Just changed the way env variables are passed in github actions -- looks like it is better to use the env parameter (which, you can really fine-tune by task) to pass in environment variables.

napulen:
Awesome, thank you @dchiller!

dchiller:
Great issues :) 

To # 1. I wonder what happens if you try to delete the data directory...or otherwise completely flush whatever is in postgres when you build? Or was this from a clean db? I know the solr re-indexes everything in the db when you re-build, so I have had the same thing happen (the app container act like it is hanging at that point, but the site still being accessible)...I don't know why that is -- on the one hand, it doesn't seem to create any issues; on the other, it might be worth looking into more?

To #2. This is definitely weird...even if the process hangs, you should see the intermediate page immediately.

dchiller:
I think the issue is now solved and as a bonus, the celery worker is now separately containerized!

napulen (approved):

PR #507, opened 2022-07-07
Changes build source for celery container

dchiller:
Previous docker compose depended on app image being built first.
Now, build order does not matter.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -65,7 +65,7 @@ services:
     restart: always
 
   celery:
-    image:  cantus_app
+    build:  ./app
     container_name: cantus-celery-1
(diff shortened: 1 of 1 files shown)


napulen:
Doesn't this mean that the image is built twice?

dchiller:
I don't think it does, looking at the logs and the created images...because of docker's cacheing mechanisms.

However, there is actually a better solution. If you are building for the first time, or you've flushed your images or docker cache, then you just need to use the command `docker compose build` first independently before you do `docker compose up`. 

And then any future time, you can use `docker compose up --build` if you want to rebuild.

So I'll just make a new PR and add that to the ReadMe.

Issue #476, opened 2022-06-07
Mapped folio shows as unmapped


eliwaksbaum:
Right now F-Pnm lat. 15181 shows the chants when you look at it, but in the admin view and the map_folios view it's listed as unmapped

eliwaksbaum:
well it eventually fixed itself. but it took a bit

dchiller:
Probably, this is because the is_mapped attribute is not switched to True until the mapping process (writing to postgres) is complete. Perhaps that is confusing? We could have some mapping in progress designation? 

How long do you think it took though? The mapping to postgres shouldn't take longer than a minute or two... (of course, we are discussing general slow speeds elsewhere that may impact)

eliwaksbaum:
I would say closer to five minutes. I think a mapping in progress would be nice, especially since mapping and then returning to the page to see "not mapped" can make you worry it didn't go through

dchiller:
Agreed

napulen:
There should be some sort of state for this already? I vaguely remember you can find out it is in progress. If not, shouldn't be difficult to implement it.

napulen:
My bad then. Ok, so the enum sounds good.

dchiller:
perfect! will do.

dchiller:
Addressed by #510

PR #510, opened 2022-07-08
Add PENDING state to manuscript mapping

dchiller:
Adds a PENDING state to the manuscript model for is_mapped,
away from a binary state. This allows for the manuscript mapped
status to show as this interim state while processing on manuscript mapping
completes. This interim status shows on the admin/map_folios page
and the admin view during this time.


dchiller:
Addressed #476.

napulen (approved):

Issue #506, opened 2022-07-07
Make link to manuscripts more prominent in the home-page and navbar


dchiller:

napulen:
More specifically 
- [ ] Modify the `home.md` markdown to include a link to the manuscripts page there
- [ ] Make the `Manuscripts` button in the header a bit more prominent with some `css` styling

dchiller:
Addressed by #512 and https://github.com/DDMAL/cantus-staticpages/pull/3

PR #512, opened 2022-07-08
Make manuscripts link in navbar more prominent

dchiller:
Makes the manuscripts link in the header navbar more prominent
by bolding text, making text bigger, and moving the to first link
in the list.


dchiller:
Addresses #506

napulen (commented):
We are just missing one commit where the submodule is updated to the latest `main`.

I can do this one, submodules are sometimes tricky.

PR #515, opened 2022-07-11
Update map folios and chant import processes

### LINKED ISSUE #450: 504 Gateway Timeouts importing chants on staging
Three times on three seperate manuscripts importing the chants in the admin view caused a 504 error. Reloading showed the chants had been loaded with the little green checkmark.

### LINKED ISSUE #473: importing chants for multiple folios at once
only the first folio selected gets the green check when returning to the admin page. could the others be loading but not displaying? or is the process getting interrupted after the first?

dchiller:
Includes these updates:

- Create a new task for each manuscript when multiple manuscripts are selected for chat import in the admin view. Older version created a single task with (potentially) multiple manuscripts.
- Remove confirmation page after kicking off chant importing in the admin view. Confirmation is now given with a green banner in the admin view.
- Remove confirmation page after submitting a folio mapping. After #510, mapping shows as pending on the main map folios page. 
- If folio mapping is pending for a manuscript, returning to the map folios page for that manuscript will no longer show a submit button -- shows a pending message instead. This will keep users from manually submitting a duplicate mapping.

These final changes should close #450 & close #473

### DIFF
--- a/app/public/cantusdata/admin/admin.py
+++ b/app/public/cantusdata/admin/admin.py
@@ -57,9 +57,8 @@ class ManuscriptAdmin(admin.ModelAdmin):
         with the selected manuscript(s)"
     )
     def load_chants(self, request, queryset):
-        manuscript_ids = [manuscript.pk for manuscript in queryset]
-        for manuscript in manuscript_ids:
-            chant_import_task.apply_async(kwargs={"manuscript_ids": [manuscript]})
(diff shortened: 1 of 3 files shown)


napulen (commented):
Simplified a bit the for loop in `admin.py`

Waiting for tests to complete.

napulen (approved):

PR #523, opened 2022-07-15
Modify folio navigation search suggestions

dchiller:


nginx/public/node/frontend/public/js/app/manuscript-detail/DivaView.js
```diff
@@ -267,6 +294,12 @@ export default Marionette.ItemView.extend({
         input.off('submit');
         input.on('submit', this.gotoInputPage);
 
+        // Rebind the go to page input focus
+        var pageSearch = this.toolbarParentObject.find(this.divaInstance.getInstanceSelector() + 'goto-page-input');
+
+        pageSearch.off('input focus');
+        pageSearch.on('input focuse', this.showPageSuggestions)
```
napulen, line 301: Is the `focuse` intentional? Doesn't seem so

napulen (approved):

PR #550, opened 2022-07-27
Reorder docker builds

dchiller:
This slight refactoring of the docker build process makes two adjustments:

1. The `app` image build process is quicker during development and minor code changes, because only (a) critical files for the build process and (b) files that are not otherwise mounted to the container are copied during the build process. Changes to the django codebase do not require an image rebuild.
2. We now mount the django codebase to the `celery` container, rather than relying on it being part of the build. This reduces the possibility that code bases in the `celery` container and `app` container go out of sync. Previously if a change was made to the django codebase and containers were not rebuilt, the update could propagate to the `app` container through the filesystem mount, but would not update in the `celery` container.

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -2,6 +2,7 @@
 FROM python:3.6.9
 COPY django-config.sh /code/
 COPY public/requirements.txt /code/public/
+COPY cantus-staticpages/ /code/cantus-staticpages/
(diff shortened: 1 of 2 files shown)


napulen (approved):
Oh, yes, I once noticed that out-of-sync volume sharing. Thank you for catching it up!

PR #553, opened 2022-07-28
Add link to CantusDB in chant detail mode

### LINKED ISSUE #547: link to Cantus DB?
My understanding of the ultimate aim of Cantus Ultimus is that it still works together with some of the information present on the Index/Database, rather than porting all of it; so if I want to see, e.g. concordances, or other chants for the same feast, etc, I would have to head to the appropriate part of the database or Index.
Probably it should be easier for the user to do this. The Cantus ID could link to the appropriate place in the Index (for example) and the chant entry could contain a link to the appropriate place in the Database. That way you don't have to go hunting in the database for information about a chant you just saw in Cantus Ultimus...

dchiller:
Closes #547

### DIFF
--- a/app/public/cantusdata/helpers/chant_importer.py
+++ b/app/public/cantusdata/helpers/chant_importer.py
@@ -68,6 +68,7 @@ class ChantImporter:
         chant.incipit = row["incipit"].strip()
         chant.full_text = row["fulltext_standardized"].strip()
         chant.volpiano = row["volpiano"].strip()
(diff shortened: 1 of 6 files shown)


napulen (commented):
Ok, so it seems the verbosity I was seeing was mostly because several files were reformatted/reindented/etc.

Waiting on confirmation for that and then we merge

napulen (approved):

PR #554, opened 2022-08-01
Refactor nginx container build

dchiller:
I just went ahead and did this because it was pretty simple and helps with increasing the speed of container rebuilds on modifications to javascript. I don't want to close the nginx container build issue because there are still some questions (eg. what's up with those errant node_modules), but this does increase speed.

### DIFF
--- a/nginx/Dockerfile
+++ b/nginx/Dockerfile
@@ -1,14 +1,15 @@
 FROM node:10 AS frontend
-COPY ./public/node /code
+COPY ./public/node/frontend/package.json /code/frontend/
 RUN npm install -g gulp
 
 WORKDIR /code/frontend
+RUN npm install
+COPY ./public/node/ /code/
 RUN wget http://www.fawe.de/volpiano/volpiano51_web.zip
 RUN unzip volpiano51_web.zip
 RUN rm volpiano51_web.zip
 RUN mv volpiano51_web/volpiano.woff ../static/fonts/volpiano.woff
-RUN npm install
-RUN gulp build --release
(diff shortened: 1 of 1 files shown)


nginx/Dockerfile
```diff
@@ -1,14 +1,15 @@
...
 WORKDIR /code/frontend
+RUN npm install
+COPY ./public/node/ /code/
 RUN wget http://www.fawe.de/volpiano/volpiano51_web.zip
 RUN unzip volpiano51_web.zip
 RUN rm volpiano51_web.zip
 RUN mv volpiano51_web/volpiano.woff ../static/fonts/volpiano.woff
-RUN npm install
-RUN gulp build --release
+RUN gulp build
```
napulen, line 12: I'm curious about the removal of the `--release` flag in `gulp build`. I thought this made the difference between minified and not-minified `js`. We want minified by default, except when debugging, right?

napulen: Also, can we remove the `npm install`?

dchiller: > Also, can we remove the `npm install`?

No, but I moved it to a different stage in the process...see new line 6.

dchiller: > I'm curious about the removal of the `--release` flag in `gulp build`. I thought this made the difference between minified and not-minified `js`. We want minified by default, except when debugging, right?

Ah, yes...that's an error...I was debugging when I committed and forgot to change it back. Will do now.

napulen (approved):

PR #596, opened 2022-08-10
Add IIIF manifest for NL-Uu 406 (3 J 7)

dchiller:


napulen (approved):

app/public/data_dumps/manifests.csv
```diff
@@ -54,3 +54,4 @@ MA Impr. 1537,https://api.digitale-sammlungen.de/iiif/presentation/v2/bsb0009035
 PL-WRu I F 401,https://glam.uni.wroc.pl/iiif/RKP_I_F_401_23465/manifest
 US-NYcub Plimpton MS 041,https://dlc.library.columbia.edu/iiif/3/presentation/10.7916/d8-v1my-eg17/manifest
 US-PRu Princeton MS. 245,https://figgy.princeton.edu/concern/scanned_resources/aa41d61e-32d2-4937-95b3-f9452d792ef9/manifest?manifest=https://figgy.princeton.edu/concern/scanned_res ources/aa41d61e-32d2-4937-95b3-f9452d792ef9/manifest
```
ahankinson, line 56: There's a space in the US-PRu MS. 245 URL -- `scanned_res ources`

ahankinson: Also you probably don't need the `?manifest=` query parameter at all, so you could remove that.

dchiller: good catch -- seems fair and confirmed that everything up to query parameter gets you to the right place. thanks!

dchiller: Resolved by #600

PR #658, opened 2022-12-09
Adjust vowel pronunciation in chant audio

### LINKED ISSUE #651: Fix audio treatment of Latin text
- [x] Pronounce dipthongs appropriately.
- [x] Fix treatment of "y"

dchiller:
Adjusts the logic determining which vowel sound to associate with chant text in cases with two vowels in the same syllable. In particular:

- "y" is pronounced [i]
- "ae" and "oe" are pronounced [ɛ]
- "Qu" + vowel is pronounced as the second vowel

The vowel combination "au" is not included because standard ecclesiastical latin pronunciation pronounces both vowels. In this case, we continue to playback that combination as [a]. 

 Closes #651.

### DIFF
--- a/nginx/public/node/frontend/public/js/app/manuscript-detail/folio/ChantRecordView.js
+++ b/nginx/public/node/frontend/public/js/app/manuscript-detail/folio/ChantRecordView.js
@@ -191,6 +191,12 @@ function get_vowel(texti) {
 	}
 	for (var x = 0; x < strLength(texti); x++) {
 		if (texti[x].toLowerCase() == "a") {
+			// Handle "ae" dipthong
+			if (x < strLength(texti) - 1){
+				if (texti[x+1].toLowerCase() == "e"){
+					return 1;
+			} 
+		}
 			return 0;
 		}
 		if (texti[x].toLowerCase() == "e") {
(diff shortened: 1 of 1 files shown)


nginx/public/node/frontend/public/js/app/manuscript-detail/folio/ChantRecordView.js
```diff
@@ -191,6 +191,12 @@ function get_vowel(texti) {
 	}
 	for (var x = 0; x < strLength(texti); x++) {
 		if (texti[x].toLowerCase() == "a") {
+			// Handle "ae" dipthong
```
jacobdgm, line 194: (trivial comment: even though everyone pronounces it "dipthong", it's diphthong

nginx/public/node/frontend/public/js/app/manuscript-detail/folio/ChantRecordView.js
```diff
@@ -191,6 +191,12 @@ function get_vowel(texti) {
...
+			if (x < strLength(texti) - 1){
```
jacobdgm, line 195: this is a place where a more descriptive variable name, and perhaps a more descriptive comment at the top of the function, could be helpful. It took me a while to figure out that `texti` is a string representing a single syllable (I think?). Could it be called `syllStr`, `syllChars` or something similar? And the function no longer "returns the number associated with first vowel in an input string" - it returns the number of the vowel the syllable should be sung with.

jacobdgm (approved):
This looks good, but I'd suggest changing a variable name and updating the comment at the top of the function.

PR #659, opened 2022-12-09
Make chant accordion headers clickable

### LINKED ISSUE #650: Make chant record header clickable
On the folio detail page, currently only the chant text is clickable to open and close the chant detail accordion. However, the entire header reacts to a mouse hover, making it seem like the whole header should respond to a click.

dchiller:
Closes #650.


jacobdgm (approved):
I'm assuming the changes in `ChantRecordView.js` are the same as in #658, and `style="display:block"` is enough to fix the issue. I likely wouldn't have arrived at this solution - I probably would have wrapped the anchor tags around the `.panel-heading` div, but hey, if it works, it works! (Are you confident that the `a` element fully fills the `div`, so that clicking on even the edge pixels of the element causes the browser to follow the link? I'd be a tad worried that there are certain pixels that trigger the hover animation but aren't actually clickable...)

dchiller:
Closing to fix commit history.

PR #660, opened 2022-12-20
Make chant accordion header clickable

### LINKED ISSUE #650: Make chant record header clickable

dchiller:
Closes #650 by:

- wrapping the entire chant accordion div in an anchor tag, opening the chant detail panel
- modifying our customized css to match styling to the older version


nginx/public/node/frontend/public/js/app/manuscript-detail/folio/chant-item.template.html
```diff
@@ -1,10 +1,10 @@
+<a data-toggle="collapse" data-parent=".accordion" href="#collapse-<%= item_id %>">
 <div class="panel-heading">
```
jacobdgm, line 2: should this block be indented?

jacobdgm (approved):

Issue #652, opened 2022-11-28
Change source/manuscript IDs to match the IDs in Cantus Database


jacobdgm:
related to [Issue 429](https://github.com/DDMAL/CantusDB/issues/429#issuecomment-1327839305) on the CantusDB github page, I figured it would be a good idea to open the issue here, where the impactful change would actually occur. Since Cantus Ultimus is more-or-less a nice user interface over top of the data on CantusDB, it would make sense to change the manuscript identifier numbers in Cantus Ultimus to match those in CantusDB. This would make it easy to link to a manuscript from CantusDB, and it also feels like a thing that would generally make integration between the two sites simpler in other situations.

(I have not looked through all of the issues that are open on this repository, so if this is a duplicate, feel free to close it)

dchiller:
@jacobdgm 

The little note I found in the current manuscript import script in CU suggests that Cantus DB implements an API -- do you know of any docs about this? I realize that your docs also might be about new cantus, and that old cantus might be different.

jacobdgm:
I'm not aware of any documentation for OldCantus's APIs. There is some documentation on their implementation in NewCantus on the [CantusDB Wiki](https://github.com/DDMAL/CantusDB/wiki/APIs) - please let me know if this can be improved in any way. And also let me know if there's additional information you would like to have provided by an API - if it's a JSON API, I should be able to add keys without breaking things, and I can also put together something new if it's needed.

dchiller:
Well, it just seems to me that an API would be the ideal way for CU to get the manuscript data from CantusDB...it makes much more sense to me to request a json document from CantusDB than script the html as we are currently doing.

dchiller:
In which case, I think there are two stages to this:

1. Modify the Cantus Ultimus manuscripts to match Cantus DB id's now -- maybe just through a data dump or a one-off json document.  
2. Modifty Cantus Ultimus's import capability to get manuscript data from New Cantus so that when it's online, we can pull new manuscripts directly from CantusDB.


Thoughts?

fujinaga:
> Well, it just seems to me that an API would be the ideal way for CU to get the manuscript data from CantusDB...it makes much more sense to me to request a json document from CantusDB than script the html as we are currently doing.

Yes, I agree.

dchiller:
@jacobdgm 

I've looked a little more in the CantusDB endpoints, and it looks like the `json-node/<source_id>/` endpoint will give us what we need.

I have two questions//comments//confirmations: 
[...]

Issue #673, opened 2023-02-02
Implement black 2023 stable styling


dchiller:
The latest release of CU's code formatter, black, implements their 2023 "stable style." This style will remain constant through the 2023 calendar year. 

Our lint test now yields an error on our current codebase. Our existing code needs to be formatted according to this new style guide, or out format testing needs to be modified to continue using the 2022 "stable style."

dchiller:
After reviewing the change log for the 2023 stable style and seeing how it would re-format files in the repo (minimally), I think it makes sense to do a one-time re-format according to the new style.

PR #683, opened 2023-03-03
Import manuscript data using CantusDB APIs

dchiller:
This PR modifies the manuscript data import process so that manuscript data is obtained (primarily) through CantusDB's API's. 

With this PR, we:
- minimize the use of CantusDB html as a source of manuscript metadata
- obtain source id's from CantusDB for all manuscripts and use those id's, rather than Django's built-in auto-incrementing id, for manuscript objects in the Cantus Ultimus database. As such, this is a partial fix for #652.

The manuscript data import process still uses html parsing to obtain a manuscript's provenance attribute, as this is not currently available through CantusDB's API. Once API changes are implemented, this reliance on html parsing should be removed and the provenance attribute should be obtained through the API.

### DIFF
--- a/app/public/cantusdata/helpers/source_importer.py
+++ b/app/public/cantusdata/helpers/source_importer.py
@@ -1,0 +1,70 @@
+from urllib import request
+import json
+
+class SourceImporter:
+    """
+    Imports sources from CantusDB by way of its APIs.
+
+    :param str cantus_db_base_url: the domain for CantusDB.
+    """
+    def __init__(self, cantus_db_base_url):
+        self.cdb_base_url = cantus_db_base_url
+
+    def request_source_ids(self):
+        """
+        Requests the IDs of all sources available in CantusDB
+        
+        :return: a list of character source ID's
+        """
+        sources_url = self.cdb_base_url + "/json-sources/"
+        sources_response = request.urlopen(sources_url).read()
+        sources_json = json.loads(sources_response)
+        source_ids = sources_json.keys()
+        return source_ids
+
+    def request_source_data(self, source_id):
+        """
+        Requests data on an individual source in CantusDB using
+        the /json-node/ endpoint.
+        
+        :param str source_id: the ID of the source in CantusDB
+        :return: a parsed json object containing source data
+        """
+        source_url = self.cdb_base_url + "/json-node/" + source_id
+        source_response = request.urlopen(source_url).read()
+        source_json = json.loads(source_response)
+        return source_json
+
+    def source_json_extractor(source_json):
+        """
+        Extracts required elements for the CU Manuscript
+        model from the API response json.
+
+        :param json source_json: parsed json response from json-node api
+        :return: a dictionary of Manuscript attributes
+        """
+        source_dict = {}
+        source_dict["id"] = source_json["vid"]
+        source_dict["name"] = source_json["title"]
+        source_dict["siglum"] = source_json["field_siglum"]["und"][0]["value"]
+        source_dict["date"] = source_json["field_date"]["und"][0]["value"]
+        source_dict["provenance"] = source_json["field_provenance"]["und"][0]["value"]
+        source_dict["description"] = source_json["field_summary"]["und"][0]["value"]
+        return source_dict
+        
+    def collect_sources(self):
+        """
+        Collects data on all sources in CantusDB for for importing
+        into Cantus Ultimus.
+
+        :return: a list of dictionaries with metadata on each source
+        """
+        source_ids = self.request_source_ids()
+        source_list = []
+        for id in source_ids:
+            source_json = self.request_source_data(id)
+            source_attrs = self.source_json_extractor(source_json)
+            source_list.append(source_attrs)
+        return source_list
+            
+            
--- a/app/public/cantusdata/helpers/source_importer.py
+++ b/app/public/cantusdata/helpers/source_importer.py
@@ -1,5 +1,30 @@
 from urllib import request
 import json
+from html.parser import HTMLParser
+
+class ProvenanceParser(HTMLParser):
+    def __init__(self):
+        super().__init__()
+        self.provenance_class = False
+        self.is_provenance = False
+        self.provenance = ""
+
+    def handle_starttag(self, tag, attrs):
+        if tag == "div":
+            if ("class","views-field views-field-field-provenance-tax") in attrs:
+                self.provenance_class = True
+        if tag == "a":
+            if self.provenance_class:
+                self.is_provenance = True
+
+    def handle_endtag(self, tag):
+        if self.is_provenance:
+            self.provenance_class = False
+            self.is_provenance = False
+    
+    def handle_data(self, data):
+        if self.is_provenance:
+            self.provenance = data
 
 class SourceImporter:
     """
@@ -22,7 +47,7 @@ class SourceImporter:
         source_ids = sources_json.keys()
         return source_ids
 
-    def request_source_data(self, source_id):
+    def source_data_download(self, source_id):
         """
         Requests data on an individual source in CantusDB using
         the /json-node/ endpoint.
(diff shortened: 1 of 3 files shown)


app/public/cantusdata/helpers/source_importer.py
```diff
@@ -0,0 +1,70 @@
...
+        
+        :param str source_id: the ID of the source in CantusDB
+        :return: a parsed json object containing source data
+        """
+        source_url = self.cdb_base_url + "/json-node/" + source_id
+        source_response = request.urlopen(source_url).read()
+        source_json = json.loads(source_response)
+        return source_json
+
+    def source_json_extractor(source_json):
```
jacobdgm: this could be more verb-like? `extract_sources_from_json`? or even `extract_sources(json)`?

jacobdgm: also, we're missing an initial `self` parameter, aren't we?

dchiller: Function names improved

app/public/cantusdata/helpers/source_importer.py
```diff
@@ -0,0 +1,124 @@
...
+        on the source provenance, but not the shorter designation that is
+        displayed publicly in CantusDB and on Cantus Ultimus). As such, this
+        method extracts this source provenance from CantusDB html. Incorporating
+        this into CantusDB's API is an open issue 
+        (see https://github.com/DDMAL/CantusDB/issues/564), so this will be
+        simplified once that is complete.
+        """
+        source_pg_url = self.cdb_base_url + "/source/" + source_id
+        source_pg_response = request.urlopen(source_pg_url).read().decode()
+        #import pdb; pdb.set_trace()
```
jacobdgm, line 77: oops!

app/public/cantusdata/helpers/source_importer.py
```diff
@@ -0,0 +1,124 @@
...
+        source_dict["provenance"] = self.download_source_provenance(source_dict["id"])
+        if source_json["field_summary"]:
+            source_dict["description"] = source_json["field_summary"]["und"][0]["value"]
+        else:
+            source_dict["description"] = ""
+        return source_dict
+            
+    def get_source_data(self, source_id):
+        """
+        Given a CantusDB, returns a dictionary of source metadata
```
jacobdgm, line 115: Should this read "Given a Source ID from CantusDB, returns..."?

app/public/cantusdata/management/commands/import_data.py
```diff
@@ -79,30 +80,41 @@ def handle(self, *args, **options):
...
+                id = source.get("id")
+                name = source.get("name")
+                cantus_url = cdb_base_url + "source/" + id
+                csv_export_url = cdb_base_url + "sites/default/files/csv/" + id + ".csv"
+                siglum = source.get("siglum")
+                date = source.get("date")
+                provenance = source.get("provenance")
+                description = source.get("description")
+                # Populating the Manuscript model
+                manuscript = Manuscript()
```
jacobdgm, line 100: Question: where does `Manuscript` come from? I don't see it among the import statements.

app/public/cantusdata/management/commands/import_data.py
```diff
@@ -79,30 +80,41 @@ def handle(self, *args, **options):
...
-            manuscript.save()
-            i += 1
+        for source_id in source_ids:
+            try:
+                source = source_importer.get_source_data(source_id)
+                # Getting the fields from the scraper
+                id = source.get("id")
+                name = source.get("name")
+                cantus_url = cdb_base_url + "source/" + id
+                csv_export_url = cdb_base_url + "sites/default/files/csv/" + id + ".csv"
```
jacobdgm, line 94: I think these would be clearer as f strings or `string.format()`. You do this in a few other places - up to you whether you want to change them.

app/public/cantusdata/management/commands/import_data.py
```diff
@@ -79,30 +80,41 @@ def handle(self, *args, **options):
...
+                manuscript = Manuscript()
+                manuscript.id = int(id)
+                manuscript.name = name
+                manuscript.cantus_url = cantus_url
+                manuscript.csv_export_url = csv_export_url
+                manuscript.siglum = siglum
+                manuscript.date = date
+                manuscript.provenance = provenance
+                manuscript.description = description
+                manuscript.is_mapped = "UNMAPPED"
```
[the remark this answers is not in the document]
jacobdgm: ah. i see line 9 in `/models/manuscript.py`. Maybe a more descriptive name for this field would be `manuscript.mapping_status`.

app/public/cantusdata/helpers/source_importer.py
```diff
@@ -0,0 +1,124 @@
...
+            if ("class","views-field views-field-field-provenance-tax") in attrs:
+                self.provenance_class = True
+        if tag == "a":
+            if self.provenance_class:
+                self.is_provenance = True
+
+    def handle_endtag(self, tag):
+        if self.is_provenance:
+            self.provenance_class = False
+            self.is_provenance = False
```
jacobdgm, line 23: I don't really understand what's happening here. What is `self.provenance_class` meant to indicate?

app/public/cantusdata/management/commands/import_data.py
dchiller: Manuscript is imported...see line 6

app/public/cantusdata/management/commands/import_data.py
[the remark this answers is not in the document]
dchiller: Yes, this has to do with the history of this field: it used to be binary, but then we needed to expand it, but didn't change the name.

You are right that another name would be better, but I don't want to do that with this PR because it would involve changing a bunch of unrelated views and templates.

app/public/cantusdata/management/commands/import_data.py
dchiller: Good point....will change

app/public/cantusdata/helpers/source_importer.py
dchiller: Yes, will change

app/public/cantusdata/helpers/source_importer.py
dchiller: `self.provenance_class` is a flag that is set to `True` when the parser reaches the element in the html document that contains the provenance data that we need to extract. 

The relevant section of the source page html looks like this:  
```
<div class="views-field views-field-field-provenance-tax">    
    <span class="views-label views-label-field-provenance-tax">Provenance: </span>    
    <strong class="field-content">
        <a href="/provenance/3666" typeof="skos:Concept" property="rdfs:label skos:prefLabel" datatype="">Eichstätt</a>
    </strong>  
</div>
```

jacobdgm:
The two most recent commits - re: pdb and f-strings - look good. My two reservations before approving:

- I still don't understand what's going on with the Provenance Parser. If you have access to the ID of the provenance, why bother with starttags and endtags and divs with particular classes when you can go to the provenance detail page and grab whatever's in the `<h1>`s?
- the linting test failed, but maybe I don't need to worry about this? Let me know if I shouldn't worry about it.

dchiller:
> * the linting check failed, but maybe I don't need to worry about this? Let me know if I shouldn't worry about it.

Yes, this is a known issue that needs to be solved (see #673). Will get to it shortly.

dchiller:
> * I still don't understand what's going on with the Provenance Parser. If you have access to the ID of the provenance, why bother with starttags and endtags and divs with particular classes when you can go to the provenance detail page and grab whatever's in the `<h1>`s?

Another way would be to go to the provenance detail page, yes. But even if I did that, I would still need to parse the html to find the provenance text....so I'm not sure how that is materially different? Or maybe I don't understand the confusion?

jacobdgm:
Okay, I think the confusion is due to my not being familiar with `html.parser` and the HTMLParser class. Looking at some documentation, I have a better sense of what this does.

So what you're doing looks fine - I just think that the parsing will be much simpler if you scrape the Provenance page rather than the source page. Currently you need two flags, `self.provenance_class` and `self.is_provenance`, because you're checking for an anchor tag within a div. Parsing the Provenance page, you'll only need a single flag, to check whether you're within the single `<h1>`.

Perhaps what's going on with the two flags would be clear to someone used to working with HTMLParser. But if I needed to make changes to this import process and you weren't around to consult, I'd be fairly lost (as we've seen!) without comments to the effect of "We're looking for the data contained within an <a> tag, which is within a <div> tag that has the following id/classes. This flag shows whether we're within the proper div. And _this_ flag shows that we're within the <a>." Maybe renaming the flags to `in_provenance_div` and `in_provenance_a` would do it (again, I don't know what the convention is)?

So, let me approve the PR - I know the import process is about to change once NewCantus replaces Old - but future developers would probably be happy if you parsed the page that has the simpler structure, or added some commentary to explain what's going on.

jacobdgm (approved):

dchiller:
> So, let me approve the PR - I know the import process is about to change once NewCantus replaces Old - but future developers would probably be happy if you parsed the page that has the simpler structure, or added some commentary to explain what's going on.

Ok, I think it's not worth the time changing the parser to a different page and testing it given that this import process will hopefully change soon anyway, but the request for comments is a point well-taken, and I think something that should be done regardless as a best practice. Will add those.

app/public/cantusdata/helpers/source_importer.py
dchiller: > Do note that you'll be looking for `/provenance/id-of-provenance/` rather than `.../id-of-source/`.

Good point!

PR #684, opened 2023-03-09
Implement Black 2023 style

### LINKED ISSUE #673: Implement black 2023 stable styling

dchiller:
Reformats python files to adhere to Black's 2023 stable style. Most of the changes resulting from the style guide update involve removing empty lines and extra parentheses. Some reformatting also occurred on files merged while the repo's formatting test was failing.

This PR also includes a commit that expands on an erroneously truncated comment in the SourceImporter class that I discovered while reviewing Black's reformatting. 

Closes #673.

### DIFF
--- a/app/public/cantusdata/management/commands/import_data.py
+++ b/app/public/cantusdata/management/commands/import_data.py
@@ -115,9 +115,7 @@ class Command(BaseCommand):
                     self.stdout.write(f"FORBIDDEN: {source_id}")
                 else:
                     self.stdout.write(f"FAILED: {source_id}")
-        self.stdout.write(
-            f"Successfully imported {i} manuscripts into database."
-        )
+        self.stdout.write(f"Successfully imported {i} manuscripts into database.")
 
     @transaction.atomic
     def import_concordance_data(self, **options):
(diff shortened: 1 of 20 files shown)


app/public/cantusdata/management/commands/import_data.py
```diff
@@ -115,9 +115,7 @@ def import_manuscript_data(self, **options):
                     self.stdout.write(f"FORBIDDEN: {source_id}")
                 else:
                     self.stdout.write(f"FAILED: {source_id}")
-        self.stdout.write(
-            f"Successfully imported {i} manuscripts into database."
-        )
+        self.stdout.write(f"Successfully imported {i} manuscripts into database.")
```
jacobdgm, line 118: I find this one curious. I guess if it's short enough to fit on a single line, it should go on a single line, as far as Black is concerned?

No need to take any action here - I just find this interesting.

jacobdgm (approved):
Good!

app/public/cantusdata/management/commands/import_data.py
dchiller: Yeah, this one stood out to me as well....I think that is what's going on here, but it doesn't seem like this would be affected by any of the new elements of the Black style? Curious, indeed

PR #623, opened 2022-08-26
Close chant on folio switch

dchiller:
Closes #619

PR #685, opened 2023-03-24
Modify open chant state management

### LINKED ISSUE #653: `chant` query parameter inclusion in URL should cause corresponding chant detail panel to open
If you navigate to a particular folio in a particular manuscript, the right folio displays (e.g. https://cantus.simssa.ca/manuscript/74/?folio%5B0%5D=001r&pageAlias=Folio%20001r). If you click on one of the chant detail accordions on the page, it causes the URL to change, with an added `&chant` query parameter appearing. (e.g. clicking on the first chant causes the accordion to open, and the URL changes to https://cantus.simssa.ca/manuscript/74/?folio%5B0%5D=001r&chant=1&pageAlias=Folio%20001r).

This behavior suggests that if one was to navigate to the page directly by entering the latter URL into the address bar, the accordion detail panel would automatically open.

Instead, pasting https://cantus.simssa.ca/manuscript/74/?folio%5B0%5D=001r&chant=1&pageAlias=Folio%20001r into the address bar and pressing enter causes the page to load without any of the detail panels opening; the URL also goes through several rapid changes, ending up on https://cantus.simssa.ca/manuscript/74/?folio%5B0%5D=001r&pageAlias=Folio%20001r, which doesn't include a `chant` query parameter.

It would be a good idea to change this for two reasons:

- First, the `chant` query parameter does not appear to be doing anything here - perhaps having the query parameter in the URL is allowing something to happen under the hood, but from the user's perspective, specifying a `chant` has no effect.
- Second, this would be useful for implementing [issue 429](https://github.com/DDMAL/CantusDB/issues/429) on the CantusDB project - if a user clicks on a "view chant on Cantus Ultimus" link a Cantus Database chant detail page, it would be ideal if we could bring them to that particular chant, rather than simply to the folio the chant is on.

dchiller:
This PR changes behaviour of chant panes on the manuscript detail page:

1. When a `chant` parameter is set in a url (eg. `/manuscript/25/?folio[0]=008v&chant=5`), the detail panel for this chant is opened on page load. This allows users to copy and share a url to a specific chant. Closes #653. This reverses the change in #623, which was setting the open chant to `null` and overriding the `chant` parameter in the url.
2. Changes where the state (open/closed) of chants for a  manuscript is saved. Previously, this state was stored in `localStorage`, which maintained these states across multiple browser sessions. This led to multiple chants being open on return visits to a manuscript without a clear reason (see #619; in fact, those chants had just been opened but not closed on previous visits). This PR causes these states to be stored in `sessionStorage`, which maintains these states in a single tab across page reloads, but flushes states across multiple tabs and new sessions. In effect, chants are "kept open" while scrolling between folios for the short term, but are "re-set" in the longer term or in new tabs.

Per @fujinaga: eventually, the decision of how long/extensively to save open chants for a manuscript (ie. within sessions/tabs or across sessions/tabs) should be made by the user. Comments in `OpenChangeState.js` are added to bring this to the attention of other developers.

### DIFF
--- a/nginx/public/node/frontend/public/js/app/manuscript-detail/folio/FolioView.js
+++ b/nginx/public/node/frontend/public/js/app/manuscript-detail/folio/FolioView.js
@@ -76,7 +76,6 @@ export default Marionette.LayoutView.extend({
         manuscriptChannel.request('set:folio', this.model.get('number'), {replaceState: true});
         folioChannel.trigger('folioLoaded');
         this.chantCollection.reset();
-        manuscriptChannel.request('set:chant', null);
         this.assignChants();
     },
 
--- a/nginx/public/node/frontend/public/js/app/objects/OpenChantState.js
+++ b/nginx/public/node/frontend/public/js/app/objects/OpenChantState.js
@@ -84,14 +84,14 @@ export default Marionette.Object.extend({
      */
     persist: function (manuscript)
     {
-        if (window.localStorage)
+        if (window.sessionStorage)
         {
-            var key = this.getLocalStorageKey(manuscript);
+            var key = this.getSessionStorageKey(manuscript);
             var data = JSON.stringify(this.manuscripts[manuscript]);
 
             try
             {
-                window.localStorage.setItem(key, data);
+                window.sessionStorage.setItem(key, data);
             }
             catch (e)
             {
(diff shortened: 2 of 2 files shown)


jacobdgm (approved):
This looks good! One thought: I don't work much with JS, but I understand that `let` is generally preferred over `var` because its scoping is more intuitive. Would this be a good opportunity to refactor the `var`s in `OpenChantState.js` in the vicinity of the changes in this PR? They all seem to be in places where `let`s should work.

Approving; you can make these additional small changes if you think they make sense.

PR #688, opened 2023-04-02
Escape hyphens in volpiano (literal) searches

### LINKED ISSUE #687: Volpiano search errors with leading hyphen
A search string that includes a leading hyphen or multiple hyphens in a Volpiano (Literal) search might be used to limit results to including complete neumes/syllables/words. Query strings with leading hyphens return no results. 

Looking at the first chant returned with a query `fg`, we would expect the query `-fg-` to return this same chant.
image
However, it does not:
image

### LINKED ISSUE #689: Single character Volpiano (Literal) searches return no results
A single note `Volpiano (Literal)` search will return no results,
image
even when some chants clearly contain the note queried
image

While the cases where a single note search is useful are probably limited, all other search types seem to return values when a search query is one character long:
image

image

image

image

image

This is probably better kept consistent throughout the various search types.

dchiller:
Fixes #687 and fixes #689. 

Hyphens are a reserved character in solr's query language as the boolean "prohibit" operator (https://solr.apache.org/guide/6_6/the-standard-query-parser.html#TheStandardQueryParser-TheBooleanOperator-). This doesn't work so well with the logic of `Volpiano (Literal)` searches, where hyphens will be included in the search string to differentiate between neumes, syllables, and words. 

This PR modifies a volpiano query by escaping hyphens in the query string before sending the query to the solr query parser. These escape characters are then removed again before the query is displayed as part of the query results. 

The PR also adjusts the filters and tokenizers in the solr schema to allow for single-character volpiano queries. Because of the way tokenizing and filtering had been configured previously, `Volpiano` searches would return results to single-character queries while `Volpiano (Literal)` searches would not.

### DIFF
--- a/nginx/public/node/frontend/public/js/app/search/chant-search/SearchInputView.js
+++ b/nginx/public/node/frontend/public/js/app/search/chant-search/SearchInputView.js
@@ -51,13 +51,19 @@ export default Marionette.ItemView.extend({
         var searchField = this.model.get('field');
         var searchInput = this.ui.searchInput.val();
         if (searchField === 'volpiano' || searchField === 'volpiano_literal'){
+            // Ensure that search input field displays a treble clef as the default
+            // search value, and replace it if user deletes it.
             if (searchInput == "" || searchInput == "1"){
                 this.ui.searchInput.val("1-");
                 searchInput = "1-";
             }
             searchInput = searchInput.replaceAll(this.invalidVolpianoRegex, "")
             this.ui.searchInput.val(searchInput)
+            // Remove the treble clef before the string is sent to solr. Volpiano
+            // searches assume treble clef.
             searchInput = searchInput.replaceAll("1-","");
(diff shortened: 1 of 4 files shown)


nginx/public/node/frontend/public/js/app/search/chant-search/SearchInputView.js
```diff
@@ -58,6 +58,7 @@ export default Marionette.ItemView.extend({
             searchInput = searchInput.replaceAll(this.invalidVolpianoRegex, "")
             this.ui.searchInput.val(searchInput)
             searchInput = searchInput.replaceAll("1-","");
+            searchInput = searchInput.replaceAll("-","\\-");
```
jacobdgm, line 61: this line and the one above it - good places for comments?

jacobdgm (approved):
Looks good, as far as I can tell!

Feel free to add comments in the noted places, or not, as you see fit.

PR #690, opened 2023-04-10
Modify concordance import process and clean import_data commant

### LINKED ISSUE #678: Harmonize concordances import process with other static data import processes
Currently, concordances are imported from hardcoded strings in `cantusdata/helpers/scrapers/concordances.py` file. This process should be harmonized with other Cantus Ultimus import procedures, rather than hardcoded in a .py file. 

We should choose one of these two other ways that data is currently imported by Cantus Ultimus:
1. **Pull the data from Cantus DB**. This would be the preferred option if concordance data is likely to be updated in Cantus DB. In this case, it would be analogous to how we treat manuscripts or chants. 
2. **Persist the data in a csv under version control**. This is similar to how we handle IIIF manifests, and is probably ok if the concordance data isn't in Cantus DB.

dchiller:
Fixes #678 by importing concordance data from a csv file in `data_dumps`. 

Also makes some adjustments for general code cleanliness in the `import_data` command (eg. using f-strings, removing unused imports, and using python's built-in csv reader for reading csv files.

### DIFF
--- a/app/public/cantusdata/management/commands/import_data.py
+++ b/app/public/cantusdata/management/commands/import_data.py
@@ -123,7 +123,9 @@ class Command(BaseCommand):
         creates a Concordance object in the database for each
         concordance.
         """
-        with open(f"{BASE_DIR}/data_dumps/concordances.csv", "r", encoding="utf-8") as file:
+        with open(
+            f"{BASE_DIR}/data_dumps/concordances.csv", "r", encoding="utf-8"
+        ) as file:
             concord_reader = csv.DictReader(file)
             concord_counter = 0
             for c in concord_reader:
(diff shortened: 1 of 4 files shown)


app/public/cantusdata/management/commands/import_data.py
```diff
@@ -119,18 +118,29 @@ def import_manuscript_data(self, **options):
...
+        """
+        Reads concordances from data_dumps/concordances.csv and
+        creates a Concordance object in the database for each
+        concordance.
+        """
+        with open(
+            f"{BASE_DIR}/data_dumps/concordances.csv", "r", encoding="utf-8"
+        ) as file:
+            concord_reader = csv.DictReader(file)
+            concord_counter = 0
```
jacobdgm, line 130: Would a more accurate variable name be `concord_count`?

jacobdgm (approved):
This looks good!

PR #702, opened 2023-05-04
Use Provenance and Siglum for page title element

### LINKED ISSUE #646: page header vs tab title
At the moment, the field in the banner header of a MS is the short form title of the manuscript (e.g. "St-Maur-des-Fossés, F-Pnm 12044") while the name that shows up as "title" in the browser tab/mouseover is the long form name, which can be quite different (in this case "Paris, Bibliotheque nationale de France" etc.). I think it would be preferable to be consistent about this, if possible (though it's hardly a top-tier issue.)

dchiller:
Closes #646.


jacobdgm (approved):
This looks good!

This might be a good place for f-strings.

Approving for now, but please make these changes.

PR #709, opened 2023-05-09
Add genre to text field in solr schema

### LINKED ISSUE #549: text search doesn't search genre
Although one can search the genre field by itself, "Search all text fields" doesn't seem to actually search genre. For antiphon it gets only 7 results, presumably from the "Feast" field:

image

dchiller:
Multiple chant fields are added to the "text" element in the solr schema so that those fields are included in an "All Text Fields" search. Previously, the genre field was not included, so the genre field was not searched in an "All Text Fields" search. This modifies the solr schema so the genre field is included in that search.

Closes #549


jacobdgm (approved):
This looks reasonable!

PR #712, opened 2023-05-09
Modify folio navigation search suggestions query

### LINKED ISSUE #564: Improve folio navigation search suggestions
As @annamorphism mentioned in #546, navigation search suggestions are not ideal, writing:

> if I type in "3" I get suggestions like 13 and 23, although it's more likely I want something in the 30s or the 300's (if they exist.)

We probably need to adjust our solr queries to improve these results.

dchiller:
The solr query used to find suggestions for the folio navigation search is modified to provide more useful results. We assume that the string typed by the user is the beginning of the folio number they wish to search (after some number of leading zeros) with a wildcard following.

For example, previously if a user typed in `4`, they would receive suggestions like `0014r` and `0034v` when those are less likely that folios like `0041r` and `0401r`.

Additionally, this commit makes some changes to the folio_set.py view requested by linting.

Closes #564

### DIFF
--- a/app/public/cantusdata/views/folio_set.py
+++ b/app/public/cantusdata/views/folio_set.py
@@ -27,10 +27,8 @@ class ManuscriptFolioSetView(APIView):
 
         if "image_uri" in kwargs:
             image_uri = kwargs["image_uri"]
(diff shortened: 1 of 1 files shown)


dchiller:
Closing this PR as a result of different approach in #728

PR #725, opened 2023-05-28
Update version of solr

### LINKED ISSUE #693: Solr container uses a unsupported solr version
The current version of solr used on Cantus Ultimus is not supported. An update may be required.

dchiller:
This PR updates the version of solr used by CU to the most recent supported version. It also introduces a number of changes to the contents of the solr container, including the simplification of configurations and the schema and the removal of unused files.

 Closes #693


ahankinson:
You can probably also remove the files in the `lang` folder since I don't think you use them in your config.

dchiller:
Thanks @ahankinson -- good point

lucasmarchd01 (approved):

PR #726, opened 2023-05-28
Update docker-compose image names to new convention

dchiller:
Convention with the latest version of docker compose is to use a hyphen as the separator in image names. This commit changes the docker-compose yaml to account for this change.

This PR also updates the repo's run-tests GitHub action to use Docker Compose v2.


jacobdgm (approved):
This looks fine to me!

PR #728, opened 2023-05-29
Change folio navigation search suggestions

dchiller:
The solr query used to find suggestions for the folio navigation search is modified to provide more useful results. We assume that the string typed by the user is the beginning of the folio number they wish to search (after some number of leading zeros) with a wildcard following.

For example, previously if a user typed in 4, they would receive suggestions like 0014r and 0034v when those are less likely that folios like 0041r and 0401r.

This PR introduces an index field that is the folio number stripped of leading zeros. This is then the field searched in order to provide suggestions.

Closes https://github.com/DDMAL/cantus/issues/564

### DIFF
--- a/app/public/cantusdata/views/folio_set.py
+++ b/app/public/cantusdata/views/folio_set.py
@@ -39,16 +39,12 @@ class ManuscriptFolioSetView(APIView):
                 score=False,
             )
             if results.numFound > 0:
-                # Initialize combined results dict with the first result
-                combined_results = results.results[0]
-                # For manuscripts with double-folio images, add the possible
-                # second folio to the combined results dict.
-                if results.numFound == 2:
-                    for field in FOLIO_FIELDS:
-                        combined_results[field] = [
-                            combined_results[field],
-                            results.results[1][field],
-                        ]
+                result_contents = results.results
+                combined_results = {}
+                for field in FOLIO_FIELDS:
+                    combined_results[field] = [
+                        result_contents[i][field] for i in range(results.numFound)
+                    ]
                 return Response(combined_results)
             if results.numFound == 0:
                 return Response({"number": None})
--- a/app/public/cantusdata/views/folio_set.py
+++ b/app/public/cantusdata/views/folio_set.py
@@ -52,9 +52,8 @@ class ManuscriptFolioSetView(APIView):
             # the manuscript that have the form:
             # [some number of leading zeros][the user-entered string][wildcard of characters].
             # Here we assume at most 3 leading zeros.
-            composed_request = f"""type:"cantusdata_folio" AND manuscript_id:{manuscript_id}
-                                 AND number:(000{query_str}* OR 00{query_str}* OR 0{query_str}*
-                                 OR {query_str}*)"""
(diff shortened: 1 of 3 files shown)


app/public/cantusdata/views/folio_set.py
```diff
@@ -45,33 +43,32 @@ def get(self, request, *args, **kwargs):
...
+        if "q" in request.GET:
+            query_str = request.GET["q"]
+            # Query for suggested folio numbers should return folios associated with
+            # the manuscript that have the form:
+            # [some number of leading zeros][the user-entered string][wildcard of characters].
+            # Here we assume at most 3 leading zeros.
+            composed_request = f"""type:"cantusdata_folio" AND manuscript_id:{manuscript_id}
+                                 AND number:(000{query_str}* OR 00{query_str}* OR 0{query_str}*
+                                 OR {query_str}*)"""
+            results = solrconn.query(
```
jacobdgm, line 57: is there a reason we're using `results` here and `result` above (like line 32)?

jacobdgm (approved):
Approving, but please do consider tweaking this based on my comments!

solr/solr/cantus_ultimus_1/conf/schema.xml
```diff
@@ -27,6 +27,9 @@
     <field name="description" type="text_general" indexed="true" stored="true" />
     <field name="public" type="boolean" default="true" indexed="true" stored="true" />
     <field name="number" type="string" indexed="true" stored="true" />
+    <!-- This field removes leading zeros from the folio number in order
+    to offer better folio navigation search suggestions -->
+    <field name="number_wo_lz" type="string" indexed="true" stored="false" />
```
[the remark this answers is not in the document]
jacobdgm: This one feels okay to me. Perhaps `number_wo_leading_zeros`?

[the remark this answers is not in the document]
dchiller: Thanks, yes, I realize `lz` is not something one would just automatically connect. I will adjust.

app/public/cantusdata/views/folio_set.py
dchiller: So, I think originally this distinction was probably intentional. The query in line 32 should only ever return one result (notice that one of our query parameters is a uri) while this query can return multiple results (it may, however, return more than one result depending on the query).

However, since we have started supporting manuscript images that contain two folios, line 32 will return two results in those cases, so this distinction is actually no longer true. I will adjust.

app/public/cantusdata/views/folio_set.py
```diff
@@ -45,33 +43,31 @@ def get(self, request, *args, **kwargs):
...
-                    fields=FOLIO_FIELDS,
-                    score=False,
-                )
-                return Response(results)
+            raise Http404("Folio set query failed.")
+        if "q" in request.GET:
+            query_str = request.GET["q"]
+            # Query for suggested folio numbers should return folios associated with
+            # the manuscript that have the form:
+            # [some number of leading zeros][the user-entered string][wildcard of characters].
```
dchiller, line 53: I realize this comment is now out of date...I will fix.

Issue #735, opened 2023-06-16
Clean up front-end dependencies


dchiller:
Currently, the `nginx` container contains many unsupported dependencies that are not properly divided into dev and production dependencies. Dependencies that are not needed should be removed and development-only dependencies should only be installed locally.

dchiller:
A first pass of removing unused dependencies is found in #738. 

Some additional dependencies might be removed when #694 is completed.

PR #738, opened 2023-06-16
Remove un-used frontend (`nginx` container) dependencies

dchiller:
The following dependencies are removed with this PR. 

- `karma` and ancillary packages and files. including `babel-istanbul` and `babel-istanbul-loader`. We used to use karma as the testing framework for our front-end. However, I have never used front-end testing with Cantus Ultimus, and it looks like the last time tests were updated was about 5 years ago. We need to do front-end testing (see #736), but we will use a supported version of a testing framework.
- `jasmine` (Note: jasmine is imported to some testing scripts that are not used. This will need to be taken care of as part of #736).
- a number of packages never referenced (`exports-loader`, `imports-loader`, `jshint-stylish`, and `node-libs-browser`)
- `mkdirp` and unused import statements (this package was introduced in https://github.com/DDMAL/cantus/commit/7936521f048c03f8a61986bee3bff24a53214c22 for bundling templates, which we no longer do)

With these removed, the frontend container builds successfully and functionality remains the same.

This begins to address #735.


jacobdgm:
Approving. But for future cleaning-up PRs like this, along the lines of the Chesterton's Fence principle, I'd appreciate a list of everything being removed (you've provided this) along with a brief summary of why they'd been introduced in the first place and why they're no longer needed (which you provided partially, in places).

I'm not sure the extent to which it's worth taking the time to track down when a library was first introduced and when the last reference to it was removed, but I think a bit of this is helpful (since comments along the lines of "seemingly unused" hardly inspire confidence!).

PR #764, opened 2023-08-10
Remove edit functions from API endpoints

### LINKED ISSUE #721: `/folios`, `/chants`, and `/concordances`, and `/browse` show API endpoint UI
At many Cantus Ultimus urls, a default view from the Django Rest Framework package is shown. These include:

- `/folios`
- `/folio/{id}`
- `/chants`
- `/chant/{id}`
- `/concordances`
- `/concordance/{id}`
- `/browse`

image

Locally, this interface allows you to make changes to the database (!) with no kind of authentication. These functions don't really seem to work on staging and production and are much "uglier" because the javascript and styling files aren't found. 

In any case, while it may be useful to have these endpoints, they certainly should not allow changes, or even look like they allow changes...

dchiller:
This PR changes the class inheritance for the `ChantList`, `ChantDetail`, `FolioList`, `FolioDetail`, `ConcordanceList`, `ConcordanceDetail`, `ManuscriptList` and `ManuscriptDetail` views. `*List` views used to inherit from Django Rest Framework's `ListCreateAPIView` which allowed for both GET and POST requests. `*Detail` views used to inherit from Django Rest Framework's `RetrieveUpdateDestroyAPIView` which allowed GET, PUT, PATCH, and DELETE requests. At this time, only GET requests are required on those endpoints, and other request types had been previously allowed without authentication or verification. The now inherit from `ListAPIView` and `RetrieveAPIView` which only provides GET requests.

This PR closes #721, where a complete inventory of Cantus Ultimus's API endpoints was also undertaken to verify which views allowed these types of edit requests.

### DIFF
--- a/app/public/cantusdata/views/chant.py
+++ b/app/public/cantusdata/views/chant.py
@@ -3,13 +3,13 @@ from cantusdata.serializers.chant import ChantSerializer
 from rest_framework import generics
 
 
-class ChantList(generics.ListCreateAPIView):
+class ChantList(generics.ListAPIView):
     model = Chant
     queryset = Chant.objects.all()
     serializer_class = ChantSerializer
 
 
-class ChantDetail(generics.RetrieveUpdateDestroyAPIView):
+class ChantDetail(generics.RetrieveAPIView):
     model = Chant
     queryset = Chant.objects.all()
     serializer_class = ChantSerializer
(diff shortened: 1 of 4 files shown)


jacobdgm (approved):
This looks reasonable!

PR #765, opened 2023-08-10
Update links to CantusDB (cantus.waterloo.ca --> cantusdatabase.org)

### LINKED ISSUE #761: Update all CantusDB links to cantusdatabase.org (NewCantus)
Links need to be updated in both the Cantus Ultimus code and in the static pages.

Resolution depends on the results of https://github.com/DDMAL/CantusDB/issues/967 because we use `/genre` in CU code.

dchiller:
This PR changes all urls for CantusDB in the Cantus Ultimus code base from cantus.waterloo.ca to cantusdatabase.org. Also uses https rather than http for all requests to CantusDB.

Although the `/genre` page redirects to `/genres` on NewCantus, this PR changes the url in `helpers/scrapers/genre.py` to go directly to `/genres` rather than through the redirect.

Closes #761 along with https://github.com/DDMAL/cantus-staticpages/pull/5

### DIFF
--- a/app/public/cantusdata/helpers/source_importer.py
+++ b/app/public/cantusdata/helpers/source_importer.py
@@ -101,7 +101,6 @@ class SourceImporter:
         simplified once that is complete.
         """
         source_pg_url = f"{self.cdb_base_url}/source/{source_id}"
-        print(source_pg_url)
         source_pg_response = request.urlopen(source_pg_url).read().decode()
         source_pg_html = ProvenanceParser()
         source_pg_html.feed(source_pg_response)
--- a/app/public/cantusdata/helpers/source_importer.py
+++ b/app/public/cantusdata/helpers/source_importer.py
@@ -101,6 +101,7 @@ class SourceImporter:
         simplified once that is complete.
         """
         source_pg_url = f"{self.cdb_base_url}/source/{source_id}"
+        print(source_pg_url)
         source_pg_response = request.urlopen(source_pg_url).read().decode()
         source_pg_html = ProvenanceParser()
         source_pg_html.feed(source_pg_response)
@@ -115,27 +116,15 @@ class SourceImporter:
         :return: a dictionary of Manuscript attributes
         """
         source_dict = {}
-        source_dict["id"] = source_json["vid"]
+        source_dict["id"] = source_json["id"]
         source_dict["name"] = source_json["title"]
-        if source_json["field_siglum"]:
-            source_dict["siglum"] = source_json["field_siglum"]["und"][0]["value"]
-        else:
-            source_dict["siglum"] = ""
-        if source_json["field_date"]:
-            source_dict["date"] = source_json["field_date"]["und"][0]["value"]
-        else:
-            source_dict["date"] = ""
+        source_dict["siglum"] = source_json.get("siglum", "")
+        source_dict["date"] = source_json.get("date", "")
         ## See documentation of the download_source_provenance method for details
         ## on these commented lines.
-        # if source_json["field_provenance"]:
-        #     source_dict["provenance"] = source_json["field_provenance"]["und"][0]["value"]
-        # else:
-        #     source_dict["provenance"] = ""
+        # source_dict["provenance"] = source_json.get("provenance","")
         source_dict["provenance"] = self.download_source_provenance(source_dict["id"])
-        if source_json["field_summary"]:
-            source_dict["description"] = source_json["field_summary"]["und"][0]["value"]
-        else:
-            source_dict["description"] = ""
+        source_dict["description"] = source_json.get("summary", "")
         return source_dict
 
     def get_source_data(self, source_id):
--- a/nginx/public/node/frontend/public/js/app/manuscript-detail/folio/ChantRecordView.js
+++ b/nginx/public/node/frontend/public/js/app/manuscript-detail/folio/ChantRecordView.js
@@ -241,7 +241,7 @@ export default Marionette.ItemView.extend({
 		var formattedVolpiano = parseVolpianoSyllables(text, volpiano);
 		this.model.set('volpiano', formattedVolpiano);
 		var cdb_uri = this.model.get('cdb_uri');
-		this.model.set({ 'cdb_link_url': 'https://cantus.uwaterloo.ca/node/' + cdb_uri });
+		this.model.set({ 'cdb_link_url': 'https://cantusdatabase.org/node/' + cdb_uri });
 	},
 	ui : {
 		volpianoSyllables: ".volpiano-syllable",
--- a/nginx/public/node/frontend/public/js/app/manuscript-detail/folio/ChantRecordView.js
+++ b/nginx/public/node/frontend/public/js/app/manuscript-detail/folio/ChantRecordView.js
@@ -244,7 +244,7 @@ export default Marionette.ItemView.extend({
 		var formattedVolpiano = parseVolpianoSyllables(text, volpiano);
 		this.model.set('volpiano', formattedVolpiano);
 		var cdb_uri = this.model.get('cdb_uri');
-		this.model.set({ 'cdb_link_url': 'https://cantusdatabase.org/node/' + cdb_uri });
+		this.model.set({ 'cdb_link_url': 'https://cantusdatabase.org/chant/' + cdb_uri });
 		manuscriptChannel.on('chantAccordion:click',this.stopChantAudio, this);
 	},
 	ui : {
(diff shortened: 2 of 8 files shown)


jacobdgm (dismissed):

app/public/cantusdata/helpers/source_importer.py
```diff
@@ -101,6 +101,7 @@ def download_source_provenance(self, source_id):
         simplified once that is complete.
         """
         source_pg_url = f"{self.cdb_base_url}/source/{source_id}"
+        print(source_pg_url)
```
jacobdgm, line 104: do we want this print statement to remain here?

nginx/public/node/frontend/public/js/app/manuscript-detail/folio/ChantRecordView.js
```diff
@@ -244,7 +244,7 @@ export default Marionette.ItemView.extend({
 		var formattedVolpiano = parseVolpianoSyllables(text, volpiano);
 		this.model.set('volpiano', formattedVolpiano);
 		var cdb_uri = this.model.get('cdb_uri');
-		this.model.set({ 'cdb_link_url': 'https://cantus.uwaterloo.ca/node/' + cdb_uri });
+		this.model.set({ 'cdb_link_url': 'https://cantusdatabase.org/node/' + cdb_uri });
```
jacobdgm, line 247: `/node/` paths are deprecated. They'll redirect to their proper Chant/Source/Sequence/etc. Detail page if they were created in OldCantus, but `/node/` will not work for any objects initially created in NewCantus. Can we use `https://cantusdatabase.org/chant/` here?

jacobdgm: Also, using a string literal would improve readability here.

jacobdgm (changes requested):
Important necessary change: we can't use `/node/` paths.

nginx/public/node/frontend/public/js/app/manuscript-detail/folio/ChantRecordView.js
dchiller: > Can we use `https://cantusdatabase.org/chant/` here?

I think so....let me write through it to make sure. Every source **not** in the Bower database will have chants associated with them, rather than sequences. Therefore, ever chant associated with a non-Bower source will be accessible at `/chant/<chant-id>`. Since Cantus Ultimus only imports the sources that are returned by `json_source_exposrt` API, it will only import non-Bower sources. --> We can use `/chant/` for any chant in Cantus Ultimus. Sound right?

jacobdgm: cantusdatabase.org/json-sources _does_ return sources from the Bower segment, so that premise doesn't work. If CU is going to have sources that are part of the Bower database, we may need to attach a flag to each source indicating whether it contains Chants or Sequences.

@annamorphism, do we expect sources from the Bower segment to end up in CU? Do any sources from Bower currently exist in CU?

dchiller: > cantusdatabase.org/json-sources _does_ return sources from the Bower segment, so that premise doesn't work

The weird thing is it doesn't look like CU does have sources from the Bower segment, and we import any source returned by `json-sources` that doesn't give us a 403 Forbidden response. We end up with 207 sources, which is exactly the number that are public in the non-Bower segment on CantusDB. (https://cantusdatabase.org/sources/?segment=4063). Maybe there is somewhere where we are filtering out the Bower sources unbeknownst to me...

dchiller: Also, some sources are both in the Bower segment and not in the Bower segment?

dchiller: Ok, I've confirmed that Bower sources aren't imported to CantusUltimus except where the source is also a Bower source, in which case the non-Bower source "version" is the one imported; however, since the CantusDB API **does** return data on those sources, it's not clear to me why they aren't imported. (This is tangential to the question of whether they **should** be imported.)

jacobdgm: > Also, some sources are both in the Bower segment and not in the Bower segment?

I don't think this is possible. Can you provide an example?

jacobdgm: > > cantusdatabase.org/json-sources _does_ return sources from the Bower segment, so that premise doesn't work
> 
> The weird thing is it doesn't look like CU does have sources from the Bower segment, and we import any source returned by `json-sources` that doesn't give us a 403 Forbidden response. We end up with 207 sources, which is exactly the number that are public in the non-Bower segment on CantusDB. (https://cantusdatabase.org/sources/?segment=4063). Maybe there is somewhere where we are filtering out the Bower sources unbeknownst to me...

Hm... I was under the impression that the sources in CU were hand-picked, based on how complete the texts/volpiano are, whether there are images, etc. But I gather now that this isn't the case....

dchiller: > > Also, some sources are both in the Bower segment and not in the Bower segment?
> 
> I don't think this is possible. Can you provide an example?

https://cantusdatabase.org/source/123779 and https://cantusdatabase.org/source/666647?

These look the "same" to me. 123779 is Bower and 666647 is not. 

I probably should have clarified that by "source" here I meant an actual manuscript source and not a source object. I have no example of a source object (with a unique ID) somehow being both.

annamorphism: That is odd! Can confirm that e.g. source 123931 (Graz 1584) is in the json-export and not the CU mansucript list....
As to Jacob's question: it will be a long while before anything Bower-related is CU-ready, though it would certainly be an interesting project. (For starters, somebody would have to go hunt down which sources have image links...and clean up the "Notes" and the notation stuff...) There is one source that exists (separately catalogued) in both segments (123779 and 666647 are the same manuscript, Einsiedeln 121), so you *could* put it in (in theory anyway) and see what that would entail.
Anyway as a project I would say this is several years out, so while it would be good to be able to add Bower sources, it can be left as "somebody can add a flag for this later" and not too much deeper than that, I think.

annamorphism: looks like @dchiller cross-posted with the same example I was digging up. The two catalogs (which is basically what /source is for CantusDB) are in different places, and look different--the Bower one is only interested in the sequences and not the mass chants, and has different stuff associated: https://cantusdatabase.org/sequence/620743 and https://cantusdatabase.org/chant/671413 both describe the same chant but one has a "feast" and the other has a "rubric".

dchiller: To sum up, @jacobdgm @annamorphism 

1. There are no imminent plans to serve Bower sources on CU, so we can go ahead and assume all chants on CU will be "chants" in CantusDB and not "sequences"...therefore we can make the offending link that started this discussion into `/chants`.
2. In the course of this discussion, we've learned that Cantus Ultimus currently imports only non-Bower sources. This is fine (per 1), but I don't know why this is when it is trying to import all sources returned in the `json-sources` API. As a separate item, I will investigate this to make sure that behaviour is well-documented and not a cause or effect of unintentional side effects.

dchiller:
@jacobdgm @annamorphism Some details from my investigations:

> cantusdatabase.org/json-sources does return sources from the Bower segment, so that premise doesn't work

This is different from OldCantus, where `json-sources` did not return sources from the Bower segment. (None of the Bower source id's I've searched for have turned up in response from cantus.uwaterloo.ca/json-sources). So this would explain why Bower sources are not currently imported into Cantus Ultimus. 

And it turns out that with the changes in this PR, we **would** start importing Bower manuscripts because of that difference. I had tested that manuscripts were being correctly imported, but I guess hadn't looked that the same number of manuscripts had been imported. 

Was that change in behaviour of `json-sources` intentional? If so, I can update our import process to include a screen on Bower sources. If not, we could make a CantusDB issue and not worry about it here.

jacobdgm (approved):

dchiller:
> I'll open this issue. Good that we caught this!

Ok... I could totally have seen the current behaviour being correct...there's nothing about the name `json-sources` that suggests it should only return CANTUS segment sources. But now new Cantus will be just like Old Cantus so that's good. Thanks for this!

PR #770, opened 2023-08-24
Handle quotation marks in search strings and adjust query result header styling

### LINKED ISSUE #766: Error on search string with open quotation marks
Searching works with strings enclosed in quotation marks
image

However, if you are typing a search string that involves quotation marks, and haven't gotten to the place in your search where you close the marks, you get a "server error" message.

image

dchiller:
This PR:

- Adds closing quotations to the end of search strings if there is an unclosed quotation in the user-entered search string. Solr errors when there are unclosed quotations (ie. an odd number of quotation marks in the query string). In this case we assume that that user is either still typing, so we close the quotation to show intermediate results, or that the user forgot to close the quotation. 
- Previously, the searched-for query string was enclosed in quotation marks. But, since the search string itself may contain quotation marks, we change the styling. 
Old style:
image
New style:
image
- Modifies how we display the query for mode searches by adding a space between modes when multiple modes are searched.
Old style:
image
New style:
image

Closes #766.

### DIFF
--- a/nginx/public/node/frontend/public/js/app/search/chant-search/SearchInputView.js
+++ b/nginx/public/node/frontend/public/js/app/search/chant-search/SearchInputView.js
@@ -66,6 +66,12 @@ export default Marionette.ItemView.extend({
             // escaped hyphens in the query string.
             searchInput = searchInput.replaceAll("-","\\-");
         }
+        // Handle quotations in text field searches. Solr errors if quotation marks
+        // are not closed. If the search string contains an odd number of quotation
+        // marks, add a quotation mark to the end of the string.
+        if (["all","feast","genre","office"].includes(searchField)){
+            (searchInput.split('"').length - 1) % 2 === 1 ? searchInput += '"' : null;
+        }
         // FIXME(wabain): While this class needs to take a SearchInput model so it can initially
         // be rendered, we're not actually updating that model here - we're just triggering
         // an event which will cause the appropriate changes to propagate. That's kind of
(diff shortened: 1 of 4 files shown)


nginx/public/node/frontend/public/js/app/search/chant-search/SearchInputView.js
```diff
@@ -66,6 +66,12 @@ export default Marionette.ItemView.extend({
             // escaped hyphens in the query string.
             searchInput = searchInput.replaceAll("-","\\-");
         }
+        // Handle quotations in text field searches. Solr errors if quotation marks
+        // are not closed. If the search string contains an odd number of quotation
+        // marks, add a quotation mark to the end of the string.
+        if (["all","feast","genre","office"].includes(searchField)){
+            (searchInput.split('"').length - 1) % 2 === 1 ? searchInput += '"' : null;
```
lucasmarchd01, line 73: Does this handle single quotation marks too, and should we be concerned about quotations within quotations?

lucasmarchd01 (approved):

nginx/public/node/frontend/public/js/app/search/chant-search/SearchInputView.js
dchiller: It just handles double marks...

The honest reason is that solr does not complain with single marks (or many other punctuation marks that you wouldn't expect in any of our text -- eg. `#` `@` `%` `&` etc). As far as I can tell, solr just strips out the single marks. 

Solr does respond to other reserved characters when entered in the search box (eg. `-` and `(`) and I do think we need a better way of making this search syntax available to the user.

In any case, single marks don't cause the error so I didn't fix them here, but it's a good point that ideally we need to figure out what's going on in these other cases.

PR #784, opened 2023-09-15
Provide csv export option for folio_set view

dchiller:
This PR:
-  adds CSVRenderer for use in Django Rest Framework responses. 
- implements the renderer at the existing `/folio-set/manuscript/<manuscript-id>` endpoint. Requests to this endpoint with a `text/csv` accept header will return a csv.

### DIFF
--- a/app/public/cantusdata/renderers/csv_renderer.py
+++ b/app/public/cantusdata/renderers/csv_renderer.py
@@ -1,0 +1,20 @@
+from typing import List
+from rest_framework.renderers import BaseRenderer
+
+
+class CSVRenderer(BaseRenderer):
+    """
+    Provides a renderer class which serializes a response to CSV.
+    The renderer assumes that the response data is a list of dictionaries
+    with a consistent set of keys (which become the header row in the csv
+    result).
+    """
+
+    media_type = "text/csv"
+    format = "csv"
+
+    def render(self, data, media_type=None, renderer_context=None):
+        headers: str = ",".join([str(key) for key in data[0].keys()])
+        rows: List[str] = [",".join([str(val) for val in row.values()]) for row in data]
+        csv_str: str = "\n".join([headers] + rows)
+        return bytes(csv_str, encoding="utf-8")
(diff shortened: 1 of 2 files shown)


jacobdgm (approved):
This all looks reasonable!

In the PR initial comment, you say "Requests to this endpoint with a `text/csv` accept header will return a csv.`

Looking at the code, I see hints that this might happen (lines 13-14 in `csv_renderer.py`), but I don't really understand how this behavior is being effected. Is it a bit of Django REST Framework magic?

dchiller:
> Looking at the code, I see hints that this might happen (lines 13-14 in `csv_renderer.py`), but I don't really understand how this behavior is being effected. Is it a bit of Django REST Framework magic?

Yes, it is ... I had to learn more about how Django REST Framework does content negotiation, and it's one of those things where it's really nice that it works, but also really opaque. See additional comments on specific lines.

app/public/cantusdata/views/folio_set.py
```diff
@@ -17,8 +17,10 @@
 
 
 class ManuscriptFolioSetView(APIView):
-    serializer_class = SearchSerializer
-    renderer_classes = (JSONRenderer,)
+    renderer_classes = (
```
dchiller, line 20: Here we define the available renderers with this view. So whereas previously we only `JSONRenderer`, behind the scenes, django rest framework will look at incoming requests and try to determine which of these two renderers should be used for the request.

app/public/cantusdata/renderers/csv_renderer.py
```diff
@@ -0,0 +1,20 @@
...
+
+class CSVRenderer(BaseRenderer):
+    """
+    Provides a renderer class which serializes a response to CSV.
+    The renderer assumes that the response data is a list of dictionaries
+    with a consistent set of keys (which become the header row in the csv
+    result).
+    """
+
+    media_type = "text/csv"
```
dchiller, line 13: Here, we say that the `CSVRenderer` should be used when the requested media type is "text/csv"...so when our `ManuscriptFolioSetView` is now determining which renderer to use, it will choose the `CSVRenderer` when the request is for "text/csv"

PR #803, opened 2023-11-20
Redirect to ManuscriptListView when manuscript is not found

### LINKED ISSUE #802: Redirect requests to old manuscript id's to `/manuscripts`
We discussed trying to determine a manuscript-to-manuscript mapping or just redirecting to the main manuscript page. It seemed like the best option would be to redirect to the main manuscript page.

dchiller:
Overrides the default `retrieve` method of the ManuscriptDetailView class so that when a manuscript with the requested ID is not found, a custom `404` page is returned that directs users to the ManuscriptListView (see below).

As mentioned in the function's comment, this is meant to mitigate cases where a link to a specific manuscript was published before manuscripts in Cantus Ultimus had stable ID's.

Closes #802 

Custom 404 Page
https://user-images.githubusercontent.com/11023634/284639619-89ed0747-a41a-46ae-82f8-a69cdda4b741.png

### DIFF
--- a/app/public/cantusdata/templates/missing_manuscript_redirect.html
+++ b/app/public/cantusdata/templates/missing_manuscript_redirect.html
@@ -6,7 +6,7 @@
     <div class="row" style="padding-top:20vh;padding-bottom:20vh">
         <div class="col-lg-12">
             <p class="text-center">This manuscript has been moved.</p>
-            <p class="text-center">You can find it on <a href="{% url 'manuscript-list' %}">the Manuscripts page</a>.</p>
+            <p class="text-center">You can find it on the <a href="{% url 'manuscript-list' %}">Manuscripts</a> page.</p>
         </div>
     </div>
 
--- a/app/public/cantusdata/templates/missing_manuscript_redirect.html
+++ b/app/public/cantusdata/templates/missing_manuscript_redirect.html
@@ -1,0 +1,25 @@
+{% extends "base.html" %}
+
+{% block body %}
+
+<div class="container">
+    <div class="row" style="padding-top:20vh;padding-bottom:20vh">
+        <div class="col-lg-12">
+            <p class="text-center">This manuscript has been moved.</p>
+            <p class="text-center">You can find it on <a href="{% url 'manuscript-list' %}">the Manuscripts page</a>.</p>
+        </div>
+    </div>
+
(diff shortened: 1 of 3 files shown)


ahankinson:
Why not do both? Return a 404 (signalling to browsers, caches, search engines, etc. that it's no longer there) but with a custom 404 page pointing to the MSS list, and letting the users know that they should update their bookmarks.

ahankinson:
Further reading: https://www.hallaminternet.com/what-are-soft-404-errors-will-they-affect-rankings/

dchiller:
Ok, this makes sense! thank you

dchiller:
image

app/public/cantusdata/views/manuscript.py
```diff
@@ -22,3 +24,21 @@ class ManuscriptDetail(generics.RetrieveAPIView):
...
     template_name = "require.html"
     renderer_classes = templated_view_renderers
+
+    def retrieve(self, request, *args, **kwargs):
+        """
+        Overrides the default retrieval method
+        to return a custom 404 page if the manuscript
+        is not found. Implemented to prevent generic
+        404 errors from links to specific manuscripts
+        made before manuscript ids were stable.
```
jacobdgm, line 34: would suggest `IDs` rather than `ids`.

app/public/cantusdata/templates/missing_manuscript_redirect.html
```diff
@@ -0,0 +1,25 @@
+{% extends "base.html" %}
+
+{% block body %}
+
+<div class="container">
+    <div class="row" style="padding-top:20vh;padding-bottom:20vh">
+        <div class="col-lg-12">
+            <p class="text-center">This manuscript has been moved.</p>
+            <p class="text-center">You can find it on <a href="{% url 'manuscript-list' %}">the Manuscripts page</a>.</p>
```
jacobdgm, line 9: should the link be applied to "Manuscripts page", or even just "Manuscripts"? Including the article feels weird to me.

Both the title and the path of the page in question use just "manuscripts". Suggesting you take the "the" and the "page" out of the link.

jacobdgm (approved):
Looks good! I suggested a couple of small tweaks which you can implement if you'd like. The one in the docstring is extremely optional.

PR #809, opened 2024-01-09
Remove concordance model and related code

### LINKED ISSUE #562: Explore including django migrations in version control
Django migrations are currently not included in version control or used in migrating between releases. 

Benefits and implications of committing migrations need to be explored, especially relating to being able to maintain staging or production databases with new releases.

### LINKED ISSUE #797: Remove dependence on CAO concordances
References to CAO Concordances were removed in Cantus DB with https://github.com/DDMAL/CantusDB/pull/431.

Now that we are using New Cantus, we need to update our code to the same.

dchiller:
CAO Concordance information has historically been imported from Cantus Database but never used in Cantus Ultimus. Now that CAO concordances are no longer present in CantusDB (see https://github.com/DDMAL/CantusDB/pull/431), we need to remove them from Cantus Ultimus. As discussed in #797, if this concordance information is ever desired in Cantus Ultimus, it would need to be re-added from Cantus Index. 

This PR:

- removes the Concordance model and references to it in other models
- removes portions of commands related to the importing of concordance information
- updates tests and test data to account for these removals
- removes references to concordances in the solr schema and search frontend

Also, historically, migrations have not been committed to version control (see #562). Migrations should be committed to version control, so we take the opportunity with this model change to do so, removing migrations from `.gitignore`

Closes #797. Closes #562.


app/public/cantusdata/fixtures/2_initial_data.json
```diff
@@ -37,19 +37,6 @@
       "model": "cantusdata.manuscript",
       "pk": 3
     },
-    {
```
jacobdgm, line 40: Curious - why does CU use fixtures, and why are they committed to the repo?

app/public/cantusdata/models/chant.py
```diff
@@ -5,9 +5,8 @@ class Chant(models.Model):
     """
     A Chant belongs to a image page (or a chant can appear
     on multiple pages?)
```
jacobdgm, line 7: at some point, it might be nice to have a definitive statement here, rather than a question :P

jacobdgm (approved):
Everything here looks fine to me. I left a couple of comments, but nothing that needs to be addressed before merging.

app/public/cantusdata/models/chant.py
dchiller: yes, i would also like to know how we will solve this problem 😂😂

app/public/cantusdata/fixtures/2_initial_data.json
dchiller: Right now, our test database (ie. the one against which our test suite is run) is populated by these fixtures.

jacobdgm: ah - got it. that makes sense.

app/public/cantusdata/models/chant.py
jacobdgm: Like, in practice, I know a chant can span multiple pages.

Here, could we say something along the lines of, "a Chant belongs to an image page. In physical manuscripts, the last chant on a page often continues onto the next page. Cantus Database associates each chant with the page on which it begins, and Cantus Ultimus follows this convention [or substitute whatever CU actually does]"? A statement like this would communicate some of the nuance of the situation and how CU handles this situation, as compared to the current statement which communicates something along the lines of "it's complicated and weird, and darned if I know what to do about it!".

PR #815, opened 2024-01-22
Add `differentiae_database` and make differentia-related fields searchable

### LINKED ISSUE #804: Provide field search for differentia data
The differentia field is currently indexed, but we provide no search interface. A new search field should be added that allows users to search both the `differentia` and `differentiae_database` fields (#805 involves adding that field).

Tagging @annamorphism for updates.

### LINKED ISSUE #805: Add `differentiae_database` to the chant model
We want to store this field on chants in the database, and index it in solr for use in search (see #804)

dchiller:
This PR:

- Adds the `differentiae_database` field from CantusDB to the chant model (Closes #805)
- Makes both the `differentia` and `differentiae_database` field searchable (Closes #804)


app/public/cantusdata/models/chant.py
```diff
@@ -25,6 +25,7 @@ class Meta:
     lit_position = models.CharField(max_length=255, blank=True, null=True)
     mode = models.CharField(max_length=255, blank=True, null=True)
     differentia = models.CharField(max_length=255, blank=True, null=True)
+    differentiae_database = models.CharField(max_length=255, blank=True)
```
ahankinson, line 28: - You should try sizing these fields to the actual expected length of the data, instead of just copying it.
- `blank=True` generally also needs `null=True`. The former validates that forms in the admin interface can be blank, while the latter allows blank values in the database. It's not very often you would need one without the other.

README.md
```diff
@@ -28,9 +28,11 @@ The build process relies on environment variables specified in an `.env` file lo
 
 Make a copy of `.env.sample` and name it `.env`. You must make two modifications to this file before the docker containers will build.  Both `POSTGRES_PASSWORD` and `RABBIT_PASSWORD` should be uncommented and set with secure passwords.
 
+Before launching the site, ensure that the `PORT` variable is set correctly. For development, it should be set to `8000`; for deployment, it should be `80`. This variable not only sets the port the server listens on, it also configures the application's debug settings. Deploying the website with `PORT=8000` would leak debugging information on a live server and must be avoided.
```
ahankinson, line 31: I'm not sure I understand this.

Port 80 is HTTP (443 is HTTPS). This is the default for a public site. Nginx, however, is your webserver that should be public-facing, and is thus proxying your Cantus application. 

Does this mean that you're running your docker on port 80? It doesn't need to (and, in fact, shouldn't -- ports < 1000 require admin privs, whereas ports > 1000 can run with their own (sandboxed) users.)

.env.sample
```diff
@@ -3,7 +3,7 @@ COMPOSE_PROJECT_NAME=cantus
 
 #Sets the DEBUG Django setting
 #PORT = 8000 for development, PORT = 80 for production
-PORT=8000
+PORT=80
```
ahankinson, line 6: See comment below... I don't think this is right.

app/public/cantusdata/models/chant.py
dchiller: > * You should try sizing these fields to the actual expected length of the data, instead of just copying it.

Sounds good!

> * `blank=True` generally also needs `null=True`. The former validates that forms in the admin interface can be blank, while the latter allows blank values in the database. It's not very often you would need one without the other.

Sounds good....for my own understanding...my impression was that since blank CharFields and TextFields are stored as `""` they wouldn't be `null`...but it is still good practice to have `null = True` for these fields?

> * `differentia` vs. `differentiae_database`... I know latin plurals, blah blah blah, but it would be good to choose one form or another.

Yes...this is what CantusDB calls them, and I'm wary of changing the names here but not there? I also find them terribly confusing and I think we have to come up with a more clear way to distinguish them/say what they are for search (conversation ongoing in #804) (although of course we can refer to them differently/provide explanatory information in the front-end). At the very least, I think this would be a good place for some help text.

README.md
dchiller: > Nginx, however, is your webserver that should be public-facing, and is thus proxying your Cantus application.

Yes...although just a note here that the nginx server on Cantus Ultimus is not itself publicly accessible (or rather, it is accessible only from a small set of IP's: the DDMAL edge router and other DDMAL staging servers).

> Does this mean that you're running your docker on port 80? It doesn't need to (and, in fact, shouldn't -- ports < 1000 require admin privs, whereas ports > 1000 can run with their own (sandboxed) users.)

The nginx Docker container is listening on port 80, and this is either mapped to the host machine's port 80 (in the case of deployed versions, accepting incoming HTTP requests from the edge router) or the host machine's port 8000 (in the case of local development). The only reason it changes is because the django settings were configured (way back in the day) to depend on this environment variable to turn or off debug, the debug toolbar, etc. It would probably be a good idea to untangle these things, and certainly when I finally get to separating dev and prod dependencies!

In any case, I'm not sure I understand exactly what you are saying may not be correct? Does the above clarify? Raise new questions and concerns?

ahankinson: I guess I'm more confused?

https://cantus.simssa.ca is on HTTPS, and the response says `Server: nginx`. This means that somewhere, some nginx is listening on 443. 

Are you saying that the path is `Internet -> nginx "edge router" (443) -> nginx (80) -> [some docker mapping] -> (gunicorn / django on 80? django on 8000?)`

Generally this would be set up as `Internet -> nginx (443) -> gunicorn (800N)`. If you have docker it might be something like `Internet -> nginx 443 -> docker 8000:8000 -> gunicorn 8000`. 

I guess what I'm saying is that if you have your edge router, you don't actually need (and, in fact, shouldn't really) be running on port 80 on your proxied services. 

Anyway, it obviously works for you, but it seems like a weird setup to me so I just thought I would clarify...

app/public/cantusdata/models/chant.py
ahankinson: > my impression was that since blank CharFields and TextFields are stored as "" they wouldn't be null...but it is still good practice to have null = True for these fields?

If a field is empty, it should be NULL. Python "truthiness" obscures this, but `""` is not the same as `None`. The ORM maps non-nullable empty fields to `""`, but NULL empty fields to `None`. 

Bugs can manifest themselves quite easily in your code, then, if you have one check that says:

```python
if model.field:
    ... do something
```

and another that says:

```python
if model.field is None:
     ... do something
```

With `null=True` then they are both the same, since it's either a value or `None`; with `null=False` they are not since they could be `None`, an empty string `""`, or a value.

dchiller: Ah, ok...this makes sense. Thanks!

README.md
dchiller: > This means that somewhere, some nginx is listening on 443.

Yes, there is an nginx on the edge router that resolves the https and then passes an http request to the Cantus Ultimus server.

Path currently is: Internet -> nginx edge router (443) -> nginx CU  (80) -> gunicorn (8001)

Are you saying you would expect the staging edge router to proxy directly to gunicorn? 

Or just that you would expect it to proxy to nginx on the CU server but on a non-80 (eg. 8000) port?

ahankinson: Yes. :-D

There's not really a need for the CU nginx, as far as i can see. The edge router is perfectly capable of proxying to gunicorn directly. Or if you want to keep the second nginx in the loop, then you can make it listen on 8000 (or any other port, really) since it doesn't really need to run on 80.

BTW, [This](https://github.com/DDMAL/cantus/blob/main/app/public/cantusdata/settings.py#L14-L15) seems like a really bizarre way of doing it, when you could just as easily define `os.environ.get("APP_DEPLOYMENT") == "DEVELOPMENT"` or something like that. But I wasn't there, so maybe there was a good reason for it?

It also looks like gunicorn is hardcoded to run on 8001. https://github.com/DDMAL/cantus/blob/main/app/django-config.sh#L8

dchiller: Gotcha. Ok, the only other thing that the CU-side nginx does is serve static files (e.g. Diva js + css, etc)...

> But I wasn't there, so maybe there was a good reason for it?

I kind of think this is one of those things where there have been a number of significant network changes over the life of CU (move to docker, move to using an edge router, etc) that this somewhat convoluted method is an artifact of that. 

I think what I'll do for the time being is remove the commits with this change and fix this in another PR.

ahankinson: Ah, that could be the reason, if your static files are hosted on the upstream. Django isn't really great for serving static files, but nginx is.

dchiller: Will address the outstanding points here in #818

nginx/public/node/frontend/public/js/app/manuscript-detail/folio/chant-record.template.html
```diff
@@ -31,6 +31,9 @@
 <% if (differentia) { %>
 <p><b>Differentia: </b> <%= differentia.replace('w', 'ω') %></p>
 <% } %>
+<% if (differentiae_database) { %>
+    <p><b>Differentiae Database: </b> <%= differentiae_database %></p>
+<% } %>
```
jacobdgm, line 36: Not required before merging, but - I appreciate the indentation in these lines! It would really help readability if such indentation was applied to the rest of the file.

(also - what templating language is this? I'm confused by the javascripty parens-with-braces for conditionals, but wrapped in `<% %>`...)

jacobdgm (approved):
I don't understand what all is necessary for Solr (and for Marionette - is it involved here?), but nothing is jumping out at me as wrong. Let this be a "weak" approval - you might want to get another set of eyes on this before merging.

nginx/public/node/frontend/public/js/app/manuscript-detail/folio/chant-record.template.html
dchiller: The template language is underscore's I believe (underscore is a dependency of marionette) `<% ... %>` executes the javascript code at `...`

https://underscorejs.org/#template

PR #819, opened 2024-01-25
Update deployment configurations and ports

### LINKED ISSUE #818: Change deployment and port configurations
Currently, debug settings are determined by the port on which the nginx container listens. It would be much clearer if this was decoupled. Also, conversation in #815 suggests that it would be better if nginx listened (behind the edge router) on a non-80 port.

For this issue:

1. Have nginx always listen on a non-80 port.
2. Configure debug with an explicit environment variable (for example, `DEVELOPMENT: True` or `DEVELOPMENT: False`
3. Update documentation for the above.

dchiller:
With this PR:

- the CU `nginx` container now always listens on the server's port `8000` since it is only serving to the edge router
- the django app's debug settings are now turned on and off with the `DEVELOPMENT` environment variable rather than the much more opaque `PORT` variable
- `makemigrations` and `migrate` commands are no longer run automatically at server start-up, since migrations are now committed to version control (introduced in #809)
- the README is updated to account for all of these changes

Closes #818.

NOTE: This PR will require a PR to the operations repo to ensure that the edge routers route requests to port 8000.


README.md
```diff
@@ -100,13 +125,15 @@ Password (again):
 Superuser created successfully.
 ```
 
-Using your admin credentials, verify that you are able to log into the admin django site, which should be located in http://localhost:8000/admin/
+Using your admin credentials, verify that you are able to log into the admin django site, which should be located at `/admin/`. 
 
 ![image](https://user-images.githubusercontent.com/7258463/101060564-8be1e500-355d-11eb-887f-5af65b50ba13.png)
 
+When initializing the site, remember to apply migrations to the database (see )
```
jacobdgm, line 132: `(see )`?

jacobdgm (approved):
There's one place where something seems to have been overlooked - other than that, this looks good to me!

README.md
dchiller: ugh...meant to fix that. Thanks for catching!

PR #826, opened 2024-02-13
Serve Gottschalk images from prod server

dchiller:
This PR updates the CU-hosted manifest for the Gottschalk reconstruction to serve images from production rather than staging. With Release 3.2-0.12.0 (https://github.com/DDMAL/cantus/releases/tag/v.3.2-0.12.0) these images are on prod, so we can serve them from prod for both staging and production purposes.


cantaloupe/interim_files/manifests/gottschalk_manifest_w_lacunae.json
```diff
@@ -209,26 +209,26 @@
...
             {
-              "@id": "https://cantus.staging.simssa.ca/iiif/lacuna.jpg",
+              "@id": "https://cantus.simssa.ca/iiif/lacuna.jpg",
               "@type": "oa:Annotation",
               "motivation": "sc:painting",
-              "on": "https://cantus.staging.simssa.ca/iiif/lacuna.jpg",
+              "on": "https://cantus.simssa.ca/iiif/lacuna.jpg",
               "resource": {
-                "@id": "https://cantus.staging.simssa.ca/iiif/lacuna.jpg/full/full/0/default.jpg",
+                "@id": "https://cantus.simssa.ca/iiif/lacuna.jpg/full/full/0/default.jpg",
```
ahankinson, line 224: While not incorrect, I generally avoid having .jpg as a part of the identifier (lacuna.jpg) component, since it can very easily be confused with the format component of the IIIF (default.jpg) to some parsers. It also is confusing since the root of the IIIF Image API URL should be the info.json responds, so  `https://cantus.simssa.ca/iiif/lacuna.jpg` should be a 303 redirect to `https://cantus.simssa.ca/iiif/lacuna.jpg/info.json`. 

Since you don't actually need this to be a URL to an actual image, then you should probably remove this and just have something like `https://cantus.simssa.ca/iiif/lacuna`.

jacobdgm (approved):
I am not at all an expert IIIF manifests, but if the change is to replace `staging.simssa.ca` with `simssa.ca` in this entire file, I trust your text editor to have not missed anything. So I'll give my approval for now, but I do think @ahankinson's suggestion makes sense, so probably update that before merging.

ahankinson:
Also, visiting `https://cantus.simssa.ca/iiif/lacuna.jpg` redirects you to `https://cantaloupe/iiif/2/lacuna.jpg/info.json` so there's something up with the server configuration that needs to be fixed.

dchiller:
The latest commit addresses Andrew's comment. I'm going to merge to make sure at least this change works, but let's continue the discussion.

ahankinson:
> There is nothing on the other end of this (ie. there is no info.json resource for the lacuna image -- or for any image we are serving)

Then it's not IIIF. The base of the URI must resolve to an info.json response -- that's how image viewers find out how many zoom levels there are, tile sizes, etc. 

Section 5: https://iiif.io/api/image/3.0/#5-image-information

> Servers MUST support requests for image information. The response is a JSON document that includes technical properties about the full image. It may also contain rights information, and services related to the image.

Canteloupe should generate the info.json automatically, so you don't need to write it by hand.

You have a canteloupe response that is looking for the info.json, but it's not configured correctly. Changing it to `https://cantus.simssa.ca/iiif/2/lacuna.jpg/info.json` almost works too, but something in the URL rewriting is adding an extra `/2/` in there.

dchiller:
> Canteloupe should generate the info.json automatically, so you don't need to write it by hand.

Ah, ok. Then yes, just a configuration issue. The extra `2` is nginx. I'm sure the redirect is a default configuration in cantaloupe.

PR #828, opened 2024-02-14
Fix nginx.conf for cantaloupe redirects

dchiller:


nginx/nginx.conf
```diff
@@ -51,6 +51,16 @@ http {
     root /code;
 
     location /iiif {
+      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+      proxy_set_header X-Forwarded-Proto $scheme;
+      proxy_set_header Host $http_host;
+     
+      proxy_redirect off;
+      
+      location /iiif/2 {
```
ahankinson, line 60: Nested locations are not that great. What is the problem you're trying to solve here?

dchiller: Cantaloupe-initiated redirects (for example, `/lacuna` -> `/lacuna/info.json`) were getting extra `/iiif/`s and `/2/`s thrown in.

Prompted by this...perhaps a regex based approach (see PR #830

PR #830, opened 2024-02-15
Update nginx.conf for cantaloupe redirects (without nested location blocks)

dchiller:
Updates `nginx.conf` to remove the need for nested location blocks. 

Requests to the Cantus Ultimus server are directed to cantaloupe if of the form `cantus[.staging].simssa.ca/iiif` or `cantus[.staging].simssa.ca/iiif/2`. The first case (sans `/2`) captures the case of uri's to the Cantus Ultimus IIIF server (for example, those we've put in the Gottschalk reconstruction's manifest). The second case (with `/2`) captures internal redirects generated by cantaloupe.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -30,6 +30,10 @@ http {
     server app:8001 fail_timeout=0;
   }
 
+  upstream cantaloupe {
+    server cantaloupe:8182;
+  }
+  
   server {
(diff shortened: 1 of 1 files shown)


jacobdgm (approved):
A review from me on a change to nginx configuration shouldn't carry much weight, but nothing looks out-of-place to me here.

PR #835, opened 2024-02-29
OMR Search: Add class for creating ngrams of information extracted from MEI

dchiller:
This PR:

- extends the `MEIParser` class rewritten in #806 to create `MEITokenizer` which divides content parsed from the MEI file into documents that can be indexed in solr
- introduces `lxml` for MEI (xml) parsing
- does some refactoring of the previously introduced `MEIParser` and tests

This concludes (hopefully) a re-writing of tools (along with #806 and #833) needed for MEI parsing.


app/Dockerfile
```diff
@@ -1,13 +1,22 @@
 # syntax=docker/dockerfile:1
+# Download and install python dependencies in a container
+FROM python:3.8.17 as dependency-install-container
```
ahankinson, line 3: If you're doing Docker, you might as well bump this up to 3.10 at least...

dchiller: Created #840

app/public/cantusdata/test/core/helpers/mei_processing/test_bounding_box_utils.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertEqual(combined_box, expected_combined_box)
+
+    def test_combine_bounding_boxes(self) -> None:
+        bounding_boxes: List[Tuple[Zone, int]] = [
+            ({"coordinates": (0, 0, 1, 1), "rotate": 0.0}, 1),
+            ({"coordinates": (1, 1, 2, 2), "rotate": 0.0}, 1),
+            ({"coordinates": (2, 1, 3, 2), "rotate": 0.0}, 2),
+            ({"coordinates": (3, 3, 4, 4), "rotate": 0.0}, 2),
+            ({"coordinates": (4, 3, 5, 6), "rotate": 0.0}, 2),
+            ({"coordinates": (8, 8, 9, 10), "rotate": 0.0}, 3),
```
jacobdgm, line 29: The last item in each tuple is the system that the bounding box belongs to, correct? And since three distinct values appear in the second position of the tuple, we end up with three different `expected_combined_boxes`?

This could be made clearer by either renaming the test to `test_combine_bounding_boxes_multiple_systems` or adding comments.

jacobdgm: Or even, would it make sense to create a `BoundingBox` type spelling out the meaning of the two values of this tuple?

jacobdgm: Also, can I suggest reordering the coordinates of the various boxes? In `combine_bounding_boxes`, you say
```
# Combine the bounding boxes for each system, in system order
combined_boxes = []
for _, boxes in sorted(system_boxes.items()):
    ...
```
but it's not immediately clear to me that `sorted` will return a list sorted in system order rather than ordered by first coordinate or something else. I'm worried that this test is meant to test that the combined bounding boxes be ordered in system order, but it's passing coincidentally since the first items of all the coordinates happen to be in system order too.

app/public/cantusdata/test/core/helpers/mei_processing/test_mei_parser.py
```diff
@@ -171,47 +185,39 @@ def test_get_contour_from_interval(self) -> None:
...
-            self.neume_component_b2,
-            self.neume_component_b2,
-            self.neume_component_b2,
+            self.nc_elem_b2,
+            self.nc_elem_b2,
+            self.nc_elem_b2,
         ]
-        neume_components_5 = [self.neume_component_d4]
+        neume_components_5 = [self.nc_elem_d4]
         with self.subTest("Analyze neume test 1"):
```
jacobdgm, line 221: could this subTest be labelled `Analyze Pes` or something similar? Similarly for the `Analyze neume test N`s below.

app/public/cantusdata/test/core/helpers/mei_processing/test_mei_tokenizer.py
```diff
@@ -0,0 +1,164 @@
...
+            min_ngram=3,
+            max_ngram=5,
+        )
+        neume_docs_3_5_grams = tokenizer_3_5.get_neume_ngram_docs()
+        neume_component_docs_3_5_grams = tokenizer_3_5.get_neume_component_ngram_docs()
+        with self.subTest("Number of ngrams"):
+            # Number of neumes in file: 117
+            # => Number of 1- and 2-grams: 117 + 116 = 233
+            # => Number of 2- and 3-grams: 116 + 115 = 231
+            # => Number of 3-, 4-, and 5-grams: 115 + 114 + 113 = 342
```
jacobdgm, line 78: Now _these_ are helpful comments! Much appreciated!

pyproject.toml
```diff
@@ -10,6 +10,7 @@ django-extensions = "3.2.3"
 django-celery-results = "2.5.1"
 djangorestframework = "3.14.0"
 gunicorn = "20.1.0"
+lxml = "5.1.0"
```
jacobdgm, line 13: cmd-F is not finding any other instances of `lxml` in this PR. Is this meant to be part of this pull request?

jacobdgm (changes requested):

pyproject.toml
dchiller: Not sure why it didn't come up for you, but:

https://github.com/DDMAL/cantus/blob/a678ae69ec88b4158f57363df0f5cf66be6e6993/app/public/cantusdata/helpers/mei_processing/mei_parser.py#L17

app/public/cantusdata/test/core/helpers/mei_processing/test_bounding_box_utils.py
dchiller: > This could be made clearer by either renaming the test to test_combine_bounding_boxes_multiple_systems or adding comments.

Ok, I've made a few changes...including by adding subtests; since we want `combine_bounding_boxes` to handle a single-system case correctly as well, it seems like the additional subtest I added makes that clear. Let me know.

> Also, can I suggest reordering the coordinates of the various boxes?

Sure...good point.

app/public/cantusdata/helpers/mei_processing/mei_parsing_types.py
```diff
@@ -0,0 +1,94 @@
...
+        defined in MEI 'zone' elements as a tuple of
+        type CoordinatesType.
+    rotate: The rotation of the zone in degrees.
+    """
+
+    coordinates: CoordinatesType
+    rotate: float
+
+
+ContourType = Literal["u", "d", "s"]
```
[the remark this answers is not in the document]
dchiller: I'm not sure a `PitchName` type is useful because we are extracting the pitch names from the xml document itself, so static type checking won't really be able to do anything with that. What we could do is run a check during the parsing that the pitch name that is extracted is valid (i.e. is one of "c", "d", "e", etc.). I guess that would be better than getting a random key error at https://github.com/DDMAL/cantus/blob/a62f5dd53d6f579556d40085118540b079ecacca/app/public/cantusdata/helpers/mei_processing/mei_parser.py#L368-L369

`NeumeType` is now added.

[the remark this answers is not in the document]
dchiller: Ok, I've added a `try`/`except` that raises a more descriptive error in this case.

app/public/cantusdata/helpers/mei_processing/mei_parser.py
```diff
@@ -407,5 +411,5 @@ def analyze_neume(
         for nc1, nc2 in zip(neume[:-1], neume[1:])
     ]
     contours: List[ContourType] = [get_contour_from_interval(i) for i in intervals]
-    neume_type: str = NEUME_GROUPS.get("".join(contours), "Compound")
+    neume_type = NEUME_GROUPS.get("".join(contours), "Compound")
```
jacobdgm, line 414: This could be annotated `neume_type: NeumeType = NEUME_GROUPS...`. The variable name makes this clear enough though - just checking that it wasn't accidentally omitted.

jacobdgm (approved):

PR #838, opened 2024-03-20
Use poetry for python dependency management

### LINKED ISSUE #552: Optimize nginx container build
We should investigate whether there is a better way to build the nginx container. I suspect that we are not taking as much advantage of docker's caching as we could -- we could have the intermediate images where most of what is installed is installed and any static file changes are closer to the end.

### LINKED ISSUE #734: Separate out development and production python dependencies
Cantus Ultimus does not currently differentiate between environments/dependencies for production and development. Some dependencies (for example, Werkzeug) are only used for development and could be removed from production.

dchiller:
This PR changes the way Cantus Ultimus handles dependencies. In the process, it closes #734.

Change introduced:
- python dependencies are managed with the `poetry` package. These are now configured in the `pyproject.toml` file and locked with the `poetry.lock` file. A `dev` dependency group is introduced that includes python dependencies useful for development when installed outside the `app` container but unnecessary for deployment; these can now be standardized across development environments. The README is updated to account for these changes.
- The build process of the `app` container has been adjusted to account for the use of `poetry`.
- Previously, python dependencies were installed as part of the `nginx` container build in order to collect Django-managed static files to a location in that container. This PR removes the need for python in the `nginx` container build process by collecting static files in the `app` container, and sharing them via a new volume with the `nginx` container. As I result, I'm going to say this closes #552. No other optimizations currently seem available.

NOTE: The `django-extensions` (and by extension, its dependency `Werkzeug`) are in fact development dependencies, but need to be installed in the `app` container. I've opened #839 to address this at a later date.


app/install-packages.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Install poetry
```
jacobdgm, line 3: This comment is not worth its line.

A comment explaining why poetry is installed a second time later in the file would be more helpful ^^

jacobdgm (approved):
I don't know enough about Poetry for my approval to carry any weight, but I've looked through it and there don't seem to be any egregious errors.

app/install-packages.sh
ahankinson: `poetry install` installs the dependencies, not poetry itself.

PR #845, opened 2024-04-11
Update MEI parsing and creation of OMR search tokens

dchiller:
This PR makes a number of changes to `helpers/mei_processing/mei_parser.py` and `helpers/mei_processing/mei_tokenizer.py` and their associated type and test files. These changes include:
- refactoring `MEITokenizer` so that we no longer return two different types of ngram documents (one on the neume level and one on the neume component level) but a single type of ngram document. This single type always includes pitch (and therefore contour and interval) information and will also include neume names if the ngram coincides with a set of complete neumes. This refactoring ensures that we can: 1. return pitch information when a neume name is queried; and 2.  we don't have multiple ngrams (one containing pitch information and one containing neume names) for the same set of pitches.
- removing empty syllables and neume from an MEI file during parsing. It seems that previous versions of MEI encoding during the OMR process could create these empty object. While this issue has been fixed, we will, at least for a little while, encounter files from before the fix. 
- modifying the dictionaries created by `MEITokenizer` to include fields required for indexing (`id` and `type`) and fields that we want to be easily available in the documents returned by Solr (`manuscript` and `folio`)

Additional refactoring includes:
- changing `neume_type` to `neume_name` (and `NeumeType` to `NeumeName`, etc.)
- adding a neume's system to the neume component objects it contains
- add a few additional development dependencies for typing and linting

### DIFF
--- a/app/public/cantusdata/helpers/mei_processing/mei_tokenizer.py
+++ b/app/public/cantusdata/helpers/mei_processing/mei_tokenizer.py
@@ -5,6 +5,7 @@ can then be indexed by a search engine (i.e. for this project, Solr).
 """
 
 from typing import List, Iterator, Any, TypedDict, Literal
+from typing_extensions import NotRequired
 from .mei_parser import MEIParser
 from .mei_parsing_types import Neume, NeumeComponent
 from .bounding_box_utils import combine_bounding_boxes, stringify_bounding_boxes
@@ -20,10 +21,16 @@ class NgramDocument(TypedDict):
     ngram_unit: The unit of the n-gram
     location: The location of the n-gram in the MEI file (MEI Zones
         converted to JSON strings according to bounding_box_utils.stringify_bounding_boxes)
+
+    The following may be part of an NgramDocument, but are not required:
+        manuscript_id: The ID of the manuscript the n-gram belongs to.
+        folio_number: The number of the folio on which the n-gram exists.
     """
 
     ngram_unit: NgramUnitType
     location: str
+    manuscript_id: NotRequired[str]
+    folio: NotRequired[str]
 
 
 class NeumeNgramDocument(NgramDocument):
@@ -51,7 +58,7 @@ class NeumeComponentNgramDocument(NgramDocument):
 
     pitch_names: str
     intervals: str
-    contours: str
+    contour: str
 
 
 def generate_ngrams(sequence: List[Any], min_n: int, max_n: int) -> Iterator[List[Any]]:
@@ -136,7 +143,7 @@ class MEITokenizer(MEIParser):
             # neume components, and not the interval and contour following
             # the last pitch in the ngram).
             intervals = [str(comp["interval"]) for comp in ngram[:-1]]
-            contours = [comp["contour"] for comp in ngram[:-1]]
+            contour = [comp["contour"] for comp in ngram[:-1]]
             bounding_boxes = [(comp["bounding_box"], neume["system"]) for comp in ngram]
             document_location = combine_bounding_boxes(bounding_boxes)
             neume_component_documents.append(
@@ -145,7 +152,7 @@ class MEITokenizer(MEIParser):
                     "location": stringify_bounding_boxes(document_location),
                     "pitch_names": pitch_names,
                     "intervals": "_".join(intervals),
-                    "contours": "_".join(contours),
+                    "contour": "_".join(contour),
                 }
             )
         return neume_component_documents
--- a/app/public/cantusdata/helpers/mei_processing/mei_tokenizer.py
+++ b/app/public/cantusdata/helpers/mei_processing/mei_tokenizer.py
@@ -5,7 +5,7 @@ can then be indexed by a search engine (i.e. for this project, Solr).
 """
 
 import uuid
-from typing import List, Tuple, Optional
+from typing import List, Tuple, Optional, Never, Union
 from .mei_parser import MEIParser
 from .mei_parsing_types import (
     Neume,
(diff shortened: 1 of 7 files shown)


app/public/cantusdata/helpers/mei_processing/mei_tokenizer.py
```diff
@@ -85,67 +64,176 @@ def __init__(self, mei_file: str, min_ngram: int, max_ngram: int) -> None:
...
-        n-grams of neume names.
+        Flatten the neumes contained in the syllables of the MEI file.
 
-        :return: A list of dictionaries containing the n-grams
-            of neume names.
+        :return: A list of neumes.
         """
-        neumes_sequence: List[Neume] = []
+        neumes: List[Neume] = []
         for syllable in self.syllables:
```
lucasmarchd01, line 75: I understand that you are removing empty syllables and neumes from MEI files before parsing, but what if the MEI file itslef is empty or has no neumes? Should this function be Optional[List[Neume]]?

app/public/cantusdata/helpers/mei_processing/mei_tokenizer.py
```diff
@@ -85,67 +64,176 @@ def __init__(self, mei_file: str, min_ngram: int, max_ngram: int) -> None:
...
+
+        :param ncs_with_sys: A list of tuples, each containing a neume component
+            and the system number of that neume component.
+        :return: An NgramDocument containing the information from the neume components.
+        """
+        pitch_names, contour, intervals = self._stringify_neume_component_data(
+            neume_components
+        )
+        zones_with_sys = [(nc["bounding_box"], nc["system"]) for nc in neume_components]
+        location = stringify_bounding_boxes(combine_bounding_boxes(zones_with_sys))
```
lucasmarchd01, line 119: Typing here?

app/public/cantusdata/helpers/mei_processing/mei_tokenizer.py
```diff
@@ -4,13 +4,13 @@
...
+from typing_extensions import NotRequired
 from .mei_parser import MEIParser
-from .mei_parsing_types import Neume, NeumeComponent
+from .mei_parsing_types import Neume, NeumeComponent, ContourType, NeumeName
 from .bounding_box_utils import combine_bounding_boxes, stringify_bounding_boxes
 
-NgramUnitType = Literal["neume", "neume_component"]
-
 
 class NgramDocument(TypedDict):
```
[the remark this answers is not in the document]
dchiller: Good point...probably!

app/public/cantusdata/helpers/mei_processing/mei_tokenizer.py
dchiller: Ah, ok .. yeah, let me check that

dchiller: Ok, I think this actually is the correct type, but it has taken a bit of research to figure out why.

Basically, `Optional[List[Neume]]` is the same as saying `neumes` could be either `List[Neume]` or `None`. But that raises a type error because `neumes` won't ever actually be `None`, because at the very least (in the case with no syllables or no neumes in the MEI file) it will be `[]`.

For similar reasons, `List[Optional[Neume]]` doesn't work, because we'll never have `None` in the list, we just may have a list of cardinality 0. 

[This post](https://stackoverflow.com/questions/76881483/whats-the-correct-way-to-type-hint-an-empty-list-as-a-literal-in-python) led me to looking into the type annotation `Never`, so I tried something here like `Union[List[Neume], List[Never]]`, but that doesn't work because `Never`, as the name suggest, means there won't ever be something added but some times/most of the time, there will be something added. Tbh, I'm kinda confused about `Never` so this paragraph might not be exactly right.

So, yeah, I wonder if `List[Neume]` is correct because even `[]` is a list of neumes (it's just a list of 0 neumes). But, there may be another, better annotation, I just can't figure out what it is.

ahankinson: If you're assigning it to `[]` then it will always be a `List`. 

> Tbh, I'm kinda confused about Never so this paragraph might not be exactly right.

To understand `Never` I would recommend learning about Functional and Statically Typed languages like Haskell or Elm. `Never` means you can use the type system to guarantee that something will never exist. If you try to do something with a value that does exist, it will be an error that you can catch at the type / compiler level, rather than at runtime. A type of `list[Never]` means that it will be a list that can't hold any values. 

If you think of an `Int` as being something like `Int 1`, `Int 2`, etc. (e.g., a type with a value of 1 or 2), and a `String` as something like `String "abc"`, `String "hello"`, then these are both types that have values. Likewise, `List[Int]` or `List[String]` are lists that accept ints or strings, i.e., `[Int 1, Int 2, Int 3]`. Likewise, even `None` is a value, so `List[Optional[Int]]` can mean `[None, Int 1, None]`. (You know it's a value because it has an index in the list.)

A `List[Never]` can never hold anything, not even `None`. So how do you add a value to a list if that value can't be expressed? You can't. Thus `x: List[Never] = []`, and if you try to `x.append(1)` the type system should flag it and say that `1` is not of type `Never`. Since you can never instantiate a value to satisfy the type system, `x` will always be an empty list. 

It's probably not what you want in this case, though. I thought I would just chime in with that. Learning a Functional Programming language is perhaps the most useful thing I did to help me write better code.

dchiller:
@lucasmarchd01 Could you give this another look? Thanks!

lucasmarchd01 (approved):

PR #848, opened 2024-04-16
Add command for indexing MEI files in solr

dchiller:
This PR contributes to the re-implementation of MEI search in CU by adding the `index_manuscript_mei` command, which, given a manuscript ID, will parse and index the MEI for that manuscript in the solr index. The PR adds associated tests. 

This PR also:
- add a submodule `production-mei-files` which tracks the [production mei files repository](https://github.com/DDMAL/production_mei_files) which will be used to manage MEI files resulting from the OMR process
- changes location of test MEI files introduce in previous OMR-related commits (e.g. #845)
- removes the old `import_mei_data` command that was used for an older OMR-search process and is superseded by the new command
- makes some other development-related changes, including:
    - adds development dependencies (stubs, etc.) to the project dependencies
    - explicitly exports Django models to conform the linting standards


app/public/cantusdata/management/commands/index_manuscript_mei.py
```diff
@@ -0,0 +1,99 @@
...
+        solr_conn = SolrConnection(settings.SOLR_SERVER)
+        manuscript_id = options["manuscript_id"][0]
+        if options.get("flush_index"):
+            self.flush_manuscript_ngrams_from_index(solr_conn, manuscript_id)
+        else:
+            folio_map: Dict[str, str] = dict(
+                Folio.objects.filter(manuscript_id=manuscript_id).values_list(
+                    "number", "image_uri"
+                )
+            )
```
lucasmarchd01, line 72: There should be an exception for the case where `Folio.objects.filter(manuscript_id=manuscript_id)` doesn't exist.

app/public/cantusdata/management/commands/index_manuscript_mei.py
```diff
@@ -0,0 +1,99 @@
...
+            manuscript_mei_path = path.join(options["mei_dir"], str(manuscript_id))
+            manuscript_mei_files = [
+                f for f in listdir(manuscript_mei_path) if f.endswith(".mei")
+            ]
```
lucasmarchd01, line 76: There should maybe be an exception here that checks if the path exists in case `--mei-dir` is formatted incorrectly.

app/public/cantusdata/management/commands/index_manuscript_mei.py
```diff
@@ -0,0 +1,99 @@
...
+            for mei_file in manuscript_mei_files:
+                folio_number: str = mei_file.split("_")[-1].split(".")[0]
+                tokenizer = MEITokenizer(
+                    path.join(manuscript_mei_path, mei_file),
+                    min_ngram=options["min_ngram"],
+                    max_ngram=options["max_ngram"],
+                )
```
lucasmarchd01, line 83: Similarly, it might make sense to check if `mei_file` string is formatted correctly to avoid indexing errors in `folio_number`.

lucasmarchd01 (changes requested):
Looks good! I think some extra exception handling would ensure that the command will always run successfully.

app/public/cantusdata/management/commands/index_manuscript_mei.py
```diff
@@ -0,0 +1,110 @@
...
+                "If this flag is set, the command will delete all existing OMR"
+                "documents for the specified manuscript before indexing the new data."
+            ),
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        solr_conn = SolrConnection(settings.SOLR_SERVER)
+        manuscript_id = options["manuscript_id"][0]
+        if options.get("flush_index"):
+            self.flush_manuscript_ngrams_from_index(solr_conn, manuscript_id)
```
ahankinson, line 66: If you `return None` here you can get rid of the long `else` block, which makes it a bit more readable

app/public/cantusdata/management/commands/index_manuscript_mei.py
```diff
@@ -0,0 +1,110 @@
...
+            if not path.exists(manuscript_mei_path):
+                raise FileNotFoundError(f"--mei-dir path does not exist.")
+            manuscript_mei_files = [
+                f for f in listdir(manuscript_mei_path) if f.endswith(".mei")
+            ]
+            if len(manuscript_mei_files) == 0:
+                raise FileNotFoundError(f"No MEI files found in {manuscript_mei_path}.")
+            for mei_file in manuscript_mei_files:
+                folio_number: str = mei_file.split("_")[-1].split(".")[0]
+                if not folio_number in folio_numbers:
```
ahankinson, line 86: I think this is the only place where you use `folio_numbers`. To me, it's just as good to do `if not folio_number in folio_map` (since the `in` operator looks at the keys anyway). In this case, `folio_numbers` is not needed at all.

app/public/cantusdata/management/commands/index_manuscript_mei.py
dchiller: And in looking at this, I realized there was a discrepancy between the arg documentation for `flush_index` and what the command actually did! Thanks!

lucasmarchd01 (approved):

PR #871, opened 2024-05-27
Implement minor build fixes

dchiller:
This PR introduces a number of relatively minor container building and configuration related fixes that I've been collecting while developing.

- Rather than randomly creating a new django secret key every time the django container is built, we set the secret key using the `.env` file. This allows us to keep the same secret key across multiple container builds and also make use of django settings for linting and type checking in development.
- There were two "versions" of the same version of `solrpy` running around. The one on pypi had a very annoying extra print statement that clogged up the logs, while the release on the `solrpy` repository did not have that print statement. This PR switches those out in the python dependencies.
- This PR adds the `psycopg` binary version to our dependencies, since we don't need to use the pure python implementation.
- In #819, we switched from using the `PORT` environment variable to signal when we were running in development mode to using a more aptly named `DEVELOPMENT` variable, but we didn't modify the rest of the `app` container build process to properly make use of this new variable. We do that here. 
- The `celery` app was using some convoluted methods to get its settings. We simplify that in [9a21998](https://github.com/DDMAL/cantus/pull/871/commits/9a2199866801b416bbf20c9bb016605934ec3663)


dchiller:
@lucasmarchd01 could you take a look?

.github/workflows/lint_and_test.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       RABBIT_USER: admin
       RABBIT_PASSWORD: supersecurepassword99
       PORT: "8000"
+      DJANGO_SECRET_KEY: verygoodandsecurekey123456123465@@123456
```
lucasmarchd01, line 21: lol

lucasmarchd01 (approved):

Issue #628, opened 2022-08-29
feature request: a better way to see chants on multiple folios


annamorphism:
It would be nice if the text/right-hand switch happened a little later than it does now.  If a chant begins at the very bottom of a folio, like this one, there is no way to follow along, since the text switches out about 1 line into the folio. (or maybe there should be a way of "pinning" the text you're looking at, in cases where the chant goes across multiple folios?)

![image](https://user-images.githubusercontent.com/67451875/187304737-73a1088f-81b9-42c8-978b-77801043e88b.png)

dchiller:
Thanks! Great idea -- I'll look into some options

dchiller:
@fujinaga @annamorphism 

I think this issue covers what we just talked about in lab meeting -- what happens when a chant goes across multiple folios.

PR #892, opened 2024-08-06
Create previously non-existent folios during mei indexing

### LINKED ISSUE #889: OMR indexing fails when we have MEI for a folio that does not exist in DB
This case occurs when the only chant "content" on a given folio is "overhang" from a previous folio (c.f. Salzinnes, A14r). In this case, we have an MEI file for that folio (because we OMR-ed and corrected the musical content on that folio), but we don't have that folio existing in the Cantus Ultimus database (because there is no chant with that folio in CantusDB).

dchiller:
Closes #889.

When chants end on a folio where no other chants begin, the folio does not exist in CantusDB but does have an MEI file. Folio A14r in Salzinnes is a good example of this: 

image

There is only one line of music on this folio -- the ending of a [chant](https://cantusdatabase.org/chant/272660) that begins on the previous folio. Thus, we have no chant in CantusDB that has folio "A14r", but we do have a MEI file for folio "A14r" from the OMR process. 

This PR modifies the index_manuscript_mei command to create a folio in such cases (we check that it's a "real" folio by making sure the mei file follows a naming convention and then create the folio if it doesn't exist). The user is alerted to this, and then must manually add the image_uri to the folio (either through the admin panel or the map folios process) and then reindex the mei. This process if further detailed in issue #891. 

This is convoluted, but a more comprehensive solution (for example, by adjusting the map folios process) seems ill-advised at the moment given upcoming changes to the Cantus Ultimus data import and indexing process to more closely couple it to CantusDB. We'll want to come up with a better solution for these cases (hence, #891) after those changes. Note that this is also relevant to longstanding #628, which we should also come back to at that time.

### DIFF
--- a/app/public/cantusdata/management/commands/index_manuscript_mei.py
+++ b/app/public/cantusdata/management/commands/index_manuscript_mei.py
@@ -95,7 +95,7 @@ class Command(BaseCommand):
                 raise ValueError(
                     f"MEI file {mei_file} does not match the expected format."
                 )
-            if not folio_number in folio_map:
+            if not folio_number in folio_map or folio_map[folio_number] == "":
                 self.stdout.write(
                     self.style.WARNING(
                         f"Folio number {folio_number} in MEI file "
--- a/app/public/cantusdata/management/commands/index_manuscript_mei.py
+++ b/app/public/cantusdata/management/commands/index_manuscript_mei.py
@@ -4,7 +4,7 @@ import re
 
 from django.core.management.base import BaseCommand, CommandParser
 from django.conf import settings
-from solr.core import SolrConnection  # type: ignore
+from solr.core import SolrConnection  # type: ignore[import-untyped]
 
 from cantusdata.helpers.mei_processing.mei_tokenizer import MEITokenizer
 from cantusdata.models.folio import Folio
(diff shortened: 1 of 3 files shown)


dchiller:
@lucasmarchd01 Could you take a look at this?

lucasmarchd01:
Oops sorry I missed this, yes!

app/public/cantusdata/management/commands/index_manuscript_mei.py
```diff
@@ -1,21 +1,30 @@
 from typing import Any, Dict
 from os import path, listdir
+import re
 
 from django.core.management.base import BaseCommand, CommandParser
 from django.conf import settings
+from solr.core import SolrConnection  # type: ignore
```
lucasmarchd01, line 7:
```suggestion
from solr.core import SolrConnection
```

lucasmarchd01 (approved):

app/public/cantusdata/management/commands/index_manuscript_mei.py
lucasmarchd01: Oh, I just didn't know this was related to the type checker, I just thought it was an erroneous comment! 

I'm not yet using type checker for CantusDB, so I should set that up. It's okay to keep as is.

dchiller: Ok! I will add the more specific ignore then, but keep the comment in! Thanks for checking.

PR #908, opened 2024-09-20
chore/build refactor: Add django-debug-toolbar and separate debug settings

dchiller:


app/public/cantusdata/debug_settings.py
```diff
@@ -0,0 +1,22 @@
+"""
+Overwrites some main settings that are used in development.
+"""
+
+from .settings import *
+
+DEBUG = is_development
+
+DEBUG_TOOLBAR_CONFIG = {
```
lucasmarchd01, line 9: You might want to add `"DISABLE_PANELS"` here and disable all the panels by default, which will save a lot of loading time. Then you can add them back as needed when you're debugging.

README.md
```diff
@@ -28,7 +28,7 @@ The build process relies on environment variables specified in an `.env` file lo
...
 
-Before launching the site, ensure that the `DEVELOPMENT` variable is set correctly. For development, it should be set to `True`; for deployment, it should be `False`. This variable configures the application's debug settings. Deploying the website with `DEVELOPMENT=True` would leak debugging information on a live server and use Django's development server rather than gunicorn and must be avoided.
+Before launching the site, ensure that the `DEVELOPMENT` and `DJANGO_SETTINGS_MODULE` variables are set correctly. For development, `DEVELOPMENT` should be set to `True`; for deployment, it should be `False`. Deploying the website with `DEVELOPMENT=True` would leak debugging information on a live server and use Django's development server rather than gunicorn and must be avoided. `DJANGO_SETTINGS_MODULE` defaults to `cantusdata.settings`, which has the `DEBUG` setting set to `False` and does not include various debugging settings. To take advantage of those debugging applications, like `django_debug_toolbar`, uncomment `DJANGO_SETTINGS_MODULE` in the `.env` file and set to `
```
lucasmarchd01, line 31:
```suggestion
Before launching the site, ensure that the `DEVELOPMENT` and `DJANGO_SETTINGS_MODULE` variables are set correctly. For development, `DEVELOPMENT` should be set to `True`; for deployment, it should be `False`. Deploying the website with `DEVELOPMENT=True` would leak debugging information on a live server and use Django's development server rather than gunicorn and must be avoided. `DJANGO_SETTINGS_MODULE` defaults to `cantusdata.settings`, which has the `DEBUG` setting set to `False` and does not include various debugging settings. To take advantage of those debugging applications, like `django_debug_toolbar`, uncomment `DJANGO_SETTINGS_MODULE` in the `.env` file and set to `cantusdata.debug_settings`.
```

lucasmarchd01 (approved):

PR #925, opened 2024-12-17
Refactor `helpers/expandr.py`

### LINKED ISSUE #900: Type and refactor the `expandr.py` helper
A smaller task towards #899 is this check and refactoring (where necessary) of the `expandr.py` helper. Breaking this out into a separate issue since it can be merged separately and before the majority of changes.

dchiller:
Closes #900.

The most significant change in this PR to `expandr.py` is the use of the `/genres` endpoint on CantusDB to get genre code -> description mappings, rather than parsing CantusDB's html.

Removes `test_import_all_data`, which previously called the manuscript import command but then did not do any testing with the results (in other words, it tested that the command ran without any python error). Since we are in the midst of a larger refactor of data import, I think we can safely get rid of this test for the time being.


app/public/cantusdata/helpers/expandr.py
```diff
@@ -1,152 +1,148 @@
...
+        genre_map: dict[str, str] = {
+            x["name"]: x["description"] for x in response.json()["genres"]
+        }
+        return genre_map
+
+    def expand_genre(self, genre_code: str) -> str:
+        """
+        Gets the genre description based on the genre code.
+        """
+        if genre_code in self.genre_data:
```
ahankinson, line 66: It's generally better to exit early. It makes the default path clearer, and removes an indent.

```py
if genre_code not in self.genre_data:
    return genre_code

description = self.genre_data[genre_code]
# some extra stuff in parentheses is showing up
paren = description.find("(")
return description[: paren - 1] if paren != -1 else description
```

PR #936, opened 2026-01-30
Upgrade Diva.js from v5.0 to v6.0.2 for IIIF Presentation 3.0 support

### LINKED ISSUE #933: Salzinnes images not loading
Salzinnes images are not loading on the website. Other manuscripts seem fine.

kyrieb-ekat:
**Summary**
Fixes #933 - Salzinnes manuscript (123723) not displaying due to IIIF Presentation API version mismatch.

Changes
- Upgraded Diva.js from v5.0 to v6.0.2 (adds IIIF Presentation 3.0 support)
- Updated webpack and gulp configs to use new Diva build structure (`build/` instead of `js/` and `css/`)
- Updated Volpiano font from v5.1 to v6.01 (font provider updated their version)
- Fixed Volpiano extraction path (zip no longer creates subdirectory)

**Testing**
- Built successfully locally with webpack/gulp
- Needs testing on staging with actual manuscript data before merge

**Related**
- Salzinnes manifest uses IIIF Presentation 3.0: https://lib.is/IE9434868/manifest
- Diva v5 only supports Presentation 2.0, causing `"can't access property 'length', e.pgs is undefined"` error
- Diva v6.0 release: https://github.com/DDMAL/diva.js/releases/tag/v6.0.2

Thanks to @ahankinson for guidance on Diva versioning and IIIF Presentation API compatibility issues.
@ahankinson - tagging you FYI in case you want to review the Diva upgrade approach, but no obligation!


annamorphism:
@yinanazhou could you prioritize this for review? The images links for Salzinnes on Cantus all go to CU, so until this is fixed people consulting Salzinnes on CantusDB are very confused...

yinanazhou (approved):
LGTM

ahankinson:
RE: Diva 7. I would park that for a while. None of the API calls are the same, and it doesn't have any highlighting ability (yet).

kyrieb-ekat:
I hadn't done so previously. I set up a full local environment and tested the manuscript viewer end-to-end. I caught several additional API incompatibilities from the v5->v6 migration that weren't visible from my earlier reviews and tests alone (method renames, the jQuery plugin init pattern, and an IIIF service type normalization issue affecting tile URL generation). These have been fixed in the latest commits (...making the PR even bigger somehow, apologies all). 

The viewer now loads correctly with images, folio navigation, and chant display working as expected. OMR search result highlighting remains non-functional as noted in the code, since the highlight plugin was removed in Diva v6 (or so I understand, given the v6 plugin system has separate files: `download.js`, `manipulation.js`, `metadata.js`, but no `highlight.js`).

Issue #937, opened 2026-03-12
put up message about missing images


annamorphism:
Until we fix the issues with Salzinnes, MS 73, etc (#933, #922) we need to have a message to alert people that their images won't load, so they aren't stuck there forever...

ahankinson:
The 500 error is not from Gallica but from the CU manifest proxy:

- https://cantus.simssa.ca/manifest-proxy/https://gallica.bnf.fr/iiif/ark:/12148/btv1b8447768b/manifest.json  -- 500 Error
- https://gallica.bnf.fr/iiif/ark:/12148/btv1b8447768b/manifest.json -- works.

PR #938, opened 2026-03-12
feat: add manuscript viewer compatibility warning systemAdd viewer_wa…

kyrieb-ekat:
…rning field to Manuscript model with UI banner support. Set warning for Salzinnes manuscript due to manifest incompatibility. Includes migration, template updates, and management command. This is in response to issue #937. In future, a more responsive version for detecting upstream versions should be implemented. Additional banner will be deployed warning about slowness for CDN-MS Medieval 0073.

Hanwen and Yinan I've tagged y'all just to get another pair of eyes on this, and because I believe it needs to be staged on the django development server and then pushed to prod from there (?).

### DIFF
--- a/app/public/cantusdata/management/commands/set_viewer_warnings.py
+++ b/app/public/cantusdata/management/commands/set_viewer_warnings.py
@@ -24,12 +24,10 @@ class Command(BaseCommand):
             salzinnes.save()
             self.stdout.write(
                 self.style.SUCCESS(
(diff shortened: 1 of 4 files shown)


ahankinson (changes requested):

kyrieb-ekat:
Accidentally merged this, though this problem gets fixes indirectly through #936. I'm going to open a separate issue and PR for adding a general template and track for when images aren't appearing.

PR #941, opened 2026-03-26
fix: update broken link

zih-syuan:
Update FRQSC link to new URL

PR #944, opened 2026-04-09
fix(nginx): update Volpiano font version to 6.01

zih-syuan:
This PR isolates the Volpiano font version update that was previously included in #941.
The broken-link fix depends on this version change to pass checks cleanly, so I split it into its own PR first.
Updated the font download source and improved the robustness of the unzipping process in the Dockerfile.


zih-syuan:
Understood. I’ll split them into separate PRs.

One complication is that the reformatting PR alone does not pass tests unless the Volpiano version is updated too. How would you like me to handle that?

Issue #953, opened 2026-04-23
`chant_importer.py` field names inconsistent with CantusDB CSV exports


kyrieb-ekat:
When importing chants via the admin interface or `import_data chants`, the 
importer may fail or silently drop data for some manuscripts. The field names 
expected by `cantusdata/helpers/chant_importer.py` do not consistently match 
the column headers present in CantusDB's CSV exports 
(`https://cantusdatabase.org/sites/default/files/csv/{id}.csv`).

**Steps to reproduce:**
1. Follow the local setup instructions in the README
2. Import a manuscript and attempt to import its chants via the admin interface
3. Observe errors or incomplete chant data depending on the manuscript

**Expected behaviour:** All fields present in the CantusDB CSV are correctly 
parsed and stored.

**Notes:** This likely requires comparing the current importer field mappings 
against a sample of real CantusDB CSV exports to identify which columns are 
missing or misnamed. The affected manuscripts and specific mismatched fields 
should be documented as part of the fix.

kyrieb-ekat:
@etosphere encountered this behavior, so any details you have please add if I missed anything. 

When I was in there earlier, I didn't actually verify the CSV column mismatch; when I tried to run chant import locally, it failed earlier with `urlopen(None)` because my fixture manuscript had no `csv_export_url` set, so I never got as far as the parsing step. I drafted the issue based on Eto's description and the field names in `chant_importer.py` but I haven't compared them against a real CantusDB CSV export. 

@liampond do you have a recent csv from your cantorales work we could check headers against?

liampond:
Just emailed you!

Issue #952, opened 2026-04-23
Manuscript.date max_length=50 causes import failure for some manuscripts


kyrieb-ekat:
When running `import_data manuscripts` locally, the import fails for manuscripts 
whose `date` field in CantusDB exceeds 50 characters. The `Manuscript` model 
currently defines:

    date = models.CharField(max_length=50, blank=True, null=True)

This silently truncates or raises a database error depending on the backend 
configuration.

**Steps to reproduce:**
1. Follow the local setup instructions in the README
2. Run `docker compose exec app python manage.py import_data manuscripts`
3. Observe failure for manuscripts with long date strings

**Expected behaviour:** Import completes for all manuscripts, or skips 
problematic records gracefully with a warning.

**Suggested fix:** Increase `max_length` to 255 (or remove the limit 
entirely with `TextField`) and generate the corresponding migration.

PR #957, opened 2026-05-10
Fix: Manuscript import failure and chant importer inconsistency with CantusDB

etosphere:
### Manuscript import (issue #952)
- Increased `Manuscript.date` max_length to allow longer date strings from CantusDB exports

### Chant import (issue #953)
- Aligned importer field names with CantusDB CSV exports (`service` instead of `office`)
- Updated genre and position expansion logic
- Removed siglum-based manuscript lookup and now require explicit manuscript context (`mobj`)

## Notes

- `ChantImporter` now explicitly requires a manuscript object (`mobj`) to be provided.
  - Previously, it attempted to resolve manuscripts from `siglum`


kyrieb-ekat (approved):
LGTM! The formatting test probably just wants you to run a python linter (we use `Black` predominantly) so it passes; squash that into a commit and then you're good to go!

PR #959, opened 2026-05-20
K8s migration

amirsadraabdollahi:
Feat: Add manifests for k8s
Fix: Change https:/ to https:// in manifest-proxy
Feat: Update dependencies


etosphere (commented):
Nice work! The migration is solid overall, and the `MIGRATION.md` plan looks great.

One follow-up (not blocking): since this PR also includes a major Django upgrade in `pyproject.toml` (`5.0.7` -> `^6.0.0`), it'd be worth running the test against Django 6 to confirm nothing broke.

PR #969, opened 2026-07-06
Isolate all Diva.js calls behind a single viewer adapter

etosphere:
Closes #963.

### DIFF
--- a/nginx/app/src/js/manuscript-detail/DivaView.js
+++ b/nginx/app/src/js/manuscript-detail/DivaView.js
@@ -3,9 +3,7 @@ import Radio from 'backbone.radio';
 import $ from 'jquery';
 import _ from "underscore";
 
-import "diva";
-
-const diva = window.Diva;
+import DivaAdapter from './DivaAdapter';
 
 import GlobalVars from '../config/GlobalVars';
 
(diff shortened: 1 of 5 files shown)


nginx/app/src/js/manuscript-detail/DivaView.js
```diff
@@ -50,71 +46,36 @@ export default Marionette.View.extend({
...
-
-            enableKeyScroll: false,
-            enableSpaceScroll: false,
-            enableCanvas: true,
-
-            objectData: '/manifest-proxy/' + manifestUrl,
-
-            blockMobileMove: false
-        };
-
```
kyrieb-ekat, line 97: Why were these removed?

kyrieb-ekat: mostly because "Download" was user-facing, and we don't seem to return to featuring it; this is also present in #970 #971 #972 #973, so I think the download button was fully dropped, which is fine if we intended that. Even if it wasn't in scope for the original tools and capabilities, offering the ability to download an image isn't a bad idea to leave.

etosphere: I didn't find the Download button in the production version. On whether to add one, Diva v7 doesn't have an original image download, but we can download the image in the "Page View" modal by "Save View" (it will download the image with the filters and transformations you set, can proceed with zero filters). We could add our own Download button as a follow-up issue.

PR #972, opened 2026-07-06
Reimplement page navigation and OMR result focus on Diva v7

etosphere:
Closes #966.

### DIFF
--- a/nginx/app/src/js/manuscript-detail/DivaBackendV7.js
+++ b/nginx/app/src/js/manuscript-detail/DivaBackendV7.js
@@ -205,14 +205,73 @@ export default class DivaBackendV7 {
         return this.currentIndex;
     }
 
+    /** Scroll to the page with the given image URI, if the viewer has it. */
+    gotoPageByURI(uri) {
+        this.viewerElement.scrollToIndex(this.pageURIs.indexOf(uri));
+    }
+
+    goToNextPage() {
+        this.viewerElement.scrollToIndex(this.currentIndex + this.pageStep());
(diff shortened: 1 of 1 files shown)


nginx/app/src/js/manuscript-detail/DivaBackendV7.js
```diff
@@ -205,14 +205,73 @@ export default class DivaBackendV7 {
         return this.currentIndex;
     }
 
+    /** Scroll to the page with the given image URI, if the viewer has it. */
+    gotoPageByURI(uri) {
+        this.viewerElement.scrollToIndex(this.pageURIs.indexOf(uri));
+    }
```
kyrieb-ekat, line 211: If uri isn't in pageURIs, indexOf returns -1 and that gets passed to `scrollToIndex`. The v6 backend's equivalent delegates to Diva's own `gotoPageByURI`, which probably handles a bad URI gracefully this one doesn't. It's fixed in [#973](https://github.com/DDMAL/mothra/issues/973) `(if (index >= 0) this.instance.goToPage(index);`), so if these land together it's moot, but flag it if [#972](https://github.com/DDMAL/cantus/pull/972) could ever ship without #973 immediately following.

etosphere: Diva v6 remains the default backend until #977, so this path is only reachable with `localStorage.divaBackend = 'v7'` in the browser's console. It shouldn't ship to users from an intermediate PR in a PR stack.

